PR #4, opened 2018-08-29
Design Improving

giacomocarrozzo:
I added 
- possibility to add empty space inside button grid
- possibility to change color to icon

I improved mobile layout


button-card.js
```diff
@@ -41,22 +44,29 @@ class ButtonCard extends LitElement {
...
         const attribute = Object.keys(cssObject)[0]
         const value = cssObject[attribute]
+        
+        if (attribute=="color") color_custom=value;
+
         card_style += `${attribute}: ${value};\n`
       })
     }
-    
+    const color_on = state.state === 'on' ? color : ( color_custom!==''? color_custom : "var(--disabled-text-color)");
```
kuuji: I'm a little bit confused by that. Why didn't you used the `color` setting in the config?

Is that because it get's overridden by the light attribute?

no-card.js
```diff
@@ -0,0 +1,24 @@
+import {
```
kuuji: That's an interesting approach. The name threw me off a little. 
I think EmptyCard might be more accurate, what do you think?

giacomocarrozzo: Yes, I just used the first name crossed my mind :D

button-card.js
giacomocarrozzo: Because you use this setting only when toggle is active, when it's not active you used only --disabled-text-color like icon's color.

PR #10, opened 2018-09-05
dynamic color options

simo878:
So this is more of a feature request, my coding is quite limited and this is my first ever PR so sorry if its not correct.
basicly this fuctionality adds the ability to set hex color option in ui-lovelace like this : 
`color: "#1affa3"`
I assumed this would also work for setting "var(--primary-color)" but it does not (only works for direct hex value) and that is the limit of my ability.

the reason for this request is so that button colors can bet set to theme variables and thus become dynamic with theme change.
Thanks.

### DIFF
--- a/button-card.js
+++ b/button-card.js
@@ -25,7 +25,7 @@ class ButtonCard extends LitElement {
 
 
   getFontColorBasedOnBackgroundColor(backgroundColor) {
-    const parsedBackgroundColor = backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i);
+    const parsedBackgroundColor = backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i) ? backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i) : this.hexToRgb(backgroundColor.substring(1));
     let fontColor = ''; // don't override by default
     if (parsedBackgroundColor) {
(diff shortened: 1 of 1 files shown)


button-card.js
```diff
@@ -25,7 +25,7 @@ class ButtonCard extends LitElement {
 
 
   getFontColorBasedOnBackgroundColor(backgroundColor) {
-    const parsedBackgroundColor = backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i);
+    const parsedBackgroundColor = backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i) ? backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i) : this.hexToRgb(backgroundColor.substring(1));
```
kuuji: I'd do 

```javascript
const parsedRgbColor= backgroundColor.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i);
const parsedBackgroundColor = parsedRgbColor ? parsedRgbColor : this.hexToRgb(backgroundColor.substring(1));
```

That way you avoid going through the regex twice. It's just a simple regex but if you have 20 buttons, it adds up.

kuuji:
That's awesome! Thanks a lot @simo878 

Let me test that tonight, I made one suggestion but it's really not a big deal.

> I assumed this would also work for setting "var(--primary-color)" but it does not (only works for direct hex value) and that is the limit of my ability.

Yeah I haven't scratched my head too hard about the css var yet. These are only evaluated by CSS / your browser. It looks like it will involve some wacky things to make that work.

PR #21, opened 2018-10-06
Added option to use entity icon

emilp333:
I wanted to have an option to defined different icons based on state but then I realized that sometimes the entity already contains the icon information as an attribute. 
If you already customized the entity with your own icon it can fetch that attribute directly from the entity instead of defining it again.
In my scenario I wanted to show different icons based on my template cover state. The template cover already have a state_open and state_close icon.

So if you add `icon: attribute` it will look for the "icon" attribute in the entity.

I´m not familiar with javascript so the code might not be optimal but it´s working. Feel free to adjust if there is a better way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Button card
 
-Lovelace Button card for your entities. 
+Lovelace Button card for your entities.
 
 ![all](examples/all.gif)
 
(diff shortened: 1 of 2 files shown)


kuuji:
Awesome! Thanks for that @emilp333 

I'll likely include that in the next release.

PR #23, opened 2018-10-08
Add option for a "Label" Button that does nothing when clicked

jxwolstenholme:
This is almost definitely not the most elegant way of doing this but in reference to:
https://github.com/kuuji/button-card/issues/13

Added new color-type to blend in with other buttons - 'label-card'

Let me know if this is worth it.


kuuji:
Thanks @jxwolstenholme ! It's been requested by few people so I'm sure it'll be appreciated. I have no strong feeling about the implementation, not worse than the empty card imo :D 

I would greatly appreciate if you could add doc changes + examples (with screenshot if possible)

jxwolstenholme:
Hopefully sorted

Issue #38, opened 2018-12-23
Change icon based on state


brianhanifin:
I used a script to toggle the state of my garage door and the color changes as expected, but I was hoping that I could also change the icon to indicate state.

Here is my example code:
```
          - type: custom:button-card
            action: service
            color_type: card
            color: rgb(66, 134, 244)
            entity: cover.garage_door
            icon: mdi:garage
            service:
              domain: script
              action: turn_on
              data:
                entity_id: script.toggle_garage_door
            show_state: true
            state:
              - value: 'closed'
                color: var(--disabled-text-color)
                icon: mdi:garage
              - value: 'closing'
                color: rgb(66, 134, 244)
                icon: mdi:garage-open
              - value: 'open'
                color: rgb(66, 134, 244)
                icon: mdi:garage-open
              - value: 'opening'
                color: rgb(66, 134, 244)
                icon: mdi:garage-open
```

nicop4:
Yes the icon is not used when defined in a state. It was broken in release 0.0.3 with the introduction of the `attribute` functionality.
I submitted a PR to the author, I hope it will be in the next release :)

nicop4:
My bad, may be the functionality was not broken but just not present :)
The PR is on its way to the next release.

PR #42, opened 2019-01-11
Fix issue #38 - use state icon if provided

nicop4:
Will use the icon of the current state if set, use main icon otherwise.
It brings back the functionality of the icon defined in a state, broken in current release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,17 +30,17 @@ Lovelace Button card for your entities.
 | ---- | ---- | ------- | --------- | -----------
 | type | string | **Required** | `custom:button-card` | Type of the card
 | entity | string | **Required** | `switch.ac` | entity_id
-| icon | string | optional | `mdi:air-conditioner` \| `attribute` | Icon to display in place of the state. If you use attribute it will fetch the icon configured on the entity.
+| icon | string | optional | `mdi:air-conditioner` \| `attribute` | Icon to display in place of the state. Will be overriden by the icon defined defined in a state (if present). If you use keywork `attribute` it will fetch the icon configured on the entity (overrides all icons defined). 
 | color_type | string | `icon` | `icon` \| `card` \| `blank-card` \| `label-card` | Color either the background of the card or the icon inside the card. Setting this to `card` enable automatic `font` and `icon` color. This allows the text/icon to be readable even if the background color is bright/dark. Additional color-type options `blank-card` and `label-card` can be used for organisation (see examples).
 | color | string | `var(--primary-text-color)` | `auto` \| `rgb(28, 128, 199)` |  Color of the icon/card when state is `on`. `auto` sets the color based on the color of a light.
 | color_off | string | `var(--disabled-text-color)` | `rgb(28, 128, 199)` |  Color of the icon/card when state is `off`.
 | size | string | `40%` | `20px` | Size of the icon. Can be percentage or pixel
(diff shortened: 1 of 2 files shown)


nicop4:
Feel free to give any feedback! 
Thanks for your custom card

kuuji:
Thanks @nicop4 I'll have a look today or tomorrow ;)

nicop4:
Sure! Doc updated.
I read it may be a little too quickly and thought it was already described ! May be it was not broken but just not present :)
Let me know if you need any more modification.

nicop4:
@kuuji when do you plan to release next version ? I noticed another bug when using state and a size in pixel: the percentage is reapplied and the icon gets smaller and smaller when you click on it (on firefox). I will create an issue and try to fix it when I have more time, but in the meantime I don't know if you prefer release soon or fix other things?

kuuji:
I was planning to release that tonight with the fix to the cdn. I'll have a look at the bug, if I can fix it quickly I will, if not I might delay the release of this PR until we have a fix.

nicop4:
I digged around the resizing issue but unfortunately I couldn't find the root cause, it is beyond my current knowledge of the framework used and the css around the button.

There is a workaround though which is to set the size of the icon in pixels instead of a percentage. In this case everything works. I don't know if there is a reason to prefer the percentage over a size in pixel. Does the icon size needs to adapt ?
One solution could be to set a default value in pixels instead of 40%. Thats not pretty, but I let you decide, may be you will be able to fix it.

Anyway, here is what I found: when the icon needs to change (by clicking the button, changing the state in the background or right-click in firefox(?!?)) the new svg icon is loaded in the DOM and gets the percentage reapplied.
As you can see on the picture of the DOM, the `ha-icon` tag has a 40% width and height and the `svg` is inside it.

image

I'm not sure whether it is a polymer or a Firefox issue, or a problem in the generated html code from the button card.

To reproduce (with the code of the PR) just set two buttons (one with an icon depending on a state) on the same entity and click on the one that has no icon redefined in a state. The other one will shrink.

```yaml
  - title: debug
    cards:
      - type: custom:button-card
        entity: light.lampe_bureau
        icon: mdi:ceiling-light
        state:
          - value: 'on'
            icon: mdi:radiator
            color: rgb(255, 129, 0)
      - type: custom:button-card
        entity: light.lampe_bureau
        icon: mdi:desk-lamp
        # size: 70px
        color: rgb(246, 220, 0)
```

Before click
image

After click
image

And actually just switching between tabs made both icons reduced...
image

Don't know what to think about it, I dont think it is related to my changes, I will test that.

kuuji:
> Does the icon size needs to adapt ?

Yes, that's mostly for mobile and portrait vs landscape modes. Which I personally use a lot on my setup.

> One solution could be to set a default value in pixels instead of 40%. Thats not pretty, but I let you decide, may be you will be able to fix it.

That means we now prevent people from using % altogether.

> Don't know what to think about it, I dont think it is related to my changes, I will test that.

Keep me posted. I agree with you though, it doesn't look related to your code. Looks like custom component bug on firefox to me. But could as well be a lit element issue. Or me not using the framework properly (I'm not really a frontend dev :D)

nicop4:
Yep, same behaviour with r0.0.3 code. The icons shrink when changing tab after click, or right click on them.
And I agree that the size shouldn't be restricted to pixels only.

I hope we will find a solution or someone that can help us, i'm not a frontend dev neither ^^

PR #45, opened 2019-01-16
Don't edit Lovelace config

bramkragten:
You should never change the Lovelace config.
This PR makes a copy of the config before changing it and adding things.

You where adding an optional parameter `state` to the config, since this is optional it could be undefined. Undefined cannot be mapped to YAML, thus breaking the YAML editors of Lovelace.

### DIFF
--- a/button-card.js
+++ b/button-card.js
@@ -102,7 +102,7 @@ class ButtonCard extends LitElement {
     const fontColor = this.getFontColorBasedOnBackgroundColor(color);
     return html`
     <style>
-    ha-icon {  
+    ha-icon {
       display: flex;
       margin: auto;
     }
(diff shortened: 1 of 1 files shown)


kuuji:
Great find @bramkragten. Thanks a lot for that.

kuuji:
@jonostanck Safari should support the spread operator. Are you using an older version?

bramkragten:
Your browser doesn't support the spread syntax, `Object.assign({}, config)` would be a replacement that is supported by all browsers. Or update your browser :-)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility

In the next version, Home Assistant will make a copy of the config before sending it to a card so this would not be strictly necessary anymore. But I would still encourage not to edit elements provided by Home Assistant.

jonostanck:
@kuuji Yeah, it's Safari 10.1.2. Unfortunately my wife is not willing to update her devices until I give her a promise that everything will work like before... and, surprise, I won't ;)
Thanks @bramkragten - I will try this solution :)

kuuji:
@jonostanck You can tell her that what doesn't work now will likely work after 🤣

PR #59, opened 2019-02-18
Color when climate is in "heat" state

BaQs:
As the button can be colored when entity is "on", this simple patch will also enable enlighting the button when thermostat is in "heat" state


kuuji:
You can already do that as is by using the state option. And it supports all the states, not just heating.

See this example => https://github.com/kuuji/button-card#configuration-with-states

PR #66, opened 2019-02-23
style

iantrich:


DavidFW1960:
Ian.... can we change the padding to 2.5% instead of 4%?
image
This is giving the same result on all my devices with 2,5%

artomka:
I also have an extra row in blank-card and label-card. 
625f34bb-8761-4e36-a232-0d661cefc630

RomRider:
`button-card-button div` is applied after whatever you put in `styles` in the yaml.
i.e. You can redefine those values anymore in the styles block:
```
padding
text-transform
font-size
font-weight
white-space
```

By doing this:
```diff
@ button-card.js:42 @
        }
        button-card-button div {
          padding: 4%;
-        text-transform: none;
-        font-size: 1.2rem;
        }
      `;
    }
```
I can redefine everything and it looks nice.

```yaml
      style:
      - font-size: 9px
      - font-family: Roboto, sans-serif
      - letter-spacing: 0.05em
```
image
image

PR #67, opened 2019-02-23
Allows coloring climate devices

BaQs:
by testing if status is "heat" and not only "on"


BaQs:
Sorry, duplicate, I just saw there is a "state" option.

https://github.com/custom-cards/button-card#configuration-with-states

PR #70, opened 2019-02-25
[WIP] Bring back paper-button and convert to TypeScript

iantrich:
Currently some issues with the resulting bundle
`Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry at http://192.168.1.2:8123/local/button-card.js:1:41080`


iantrich:
Need to determine how to exclude Polymer from the final bundle. @balloob provided this webpack example:
https://github.com/home-assistant/home-assistant-polymer/blob/dev/config/webpack.js#L19-L24

balloob:
Generally you want to avoid making web components that have no check around customElements.define a dependency. It means that if another card would include `paper-button`, you're now going to blow up loading your whole component, as an exception is raised within your bundle.

Better to make a standalone package for paper-button that people can install, that way it won't take down your bundle if it's already defined.

But really, the solution here is to not use `<paper-button>` at all? You're overriding so much styling, there is really no reason to use it at all. Just remove it?

iantrich:
You suggest just using a good ole fashion button and let people add styles to their heart's content?

balloob:
Yeah, just use: https://gist.github.com/MoOx/9137295

iantrich:
Cool. This bundling has been a learning experience for me :) But this makes it a lot more straight-forward. Thanks!

RomRider:
@iantrich, is this still a thing with the latest refactor?

iantrich:
What do you mean?

RomRider:
Can we close it?

PR #84, opened 2019-04-16
tap_action instead of action, deprecation of color_off and style in state support

### LINKED ISSUE #31: Button as a link
Feature Request
Can the feature that a button can work as a link to a specific site be implemented? 
Like:
action: navigate
navigation_path: /lovelace/0

Like the function you have in Picture Elements Card.
Thx for the good work!

### LINKED ISSUE #41: Add support for RGBA
Hi ! it could be really cool if we could play with the transparency. so have a fancy background and set the buttons with a 50% transparent when off for exemple

### LINKED ISSUE #53: Icon from sensor not working
Hi,

If I either don't set the icon to any value or set it to `attribute` as stated it doesn't show up at all.

Icons set manually:
image

No icon set manually:
image

### LINKED ISSUE #71: Custom Style attributes break when color_type is set to Icon
Running the latest 0.0.6 release with HASS 0.88.1. Title pretty much says it all. Style attributes work perfectly when the color type is set to card, but break when I change the color_type to icon. I'm only doing some text transformation:
               - font-size: 10px
               - text-transform: capitalize

### LINKED ISSUE #75: Add service for navigation
Would it be possible to add in the navigation service so on button click it can go to a new page?

Thanks.

RomRider:
**Breaking Changes**
* `color_off` is now deprecated, use `state` instead if you want to redefine the color when `off`
  Instead of:
  ```yaml
    - type: "custom:button-card"
      name: Switch
      entity: switch.main
      color: red
      color_off: green
  ```
  Use:
  ```yaml
    - type: "custom:button-card"
      name: Switch
      entity: switch.main
      color: red
      state:
        - value: 'off'
          color: green
  ```
  Or:
  ```yaml
    - type: "custom:button-card"
      name: Switch
      entity: switch.main
      state:
        - value: 'off'
          color: green
        - value: 'on'
          color: red
  ```
* `default_color` is deprecated as it seems redundant with `color` (when light is not rgb). For rgb light, you can still define a specific color for the `off` state for example.
* `service` has been deprecated in favor of `tap_action`, see below
* `action` has been replaced by `tap_action` to mimic the default button card and allow for future options like `hold_action`
* `more_info` renamed to `more-info`, `more_info` still works
* `service` now takes the full service name `light.turn_on`
* Add default entity icon support, Fix #53, add `show_icon` to decide whether to display the icon or not

**Others**
* blink animation support, see examples
blink-animation
* `tap_action` supports `navigate`. Fix #75 & Fix #31 
* Support for style definition in state. Fix #41 
* Disable click when not needed
* Refactored the code for faster rendering
* Documentation update
* Fix #71


RomRider:
Would fix #41 also with this feature `Support for style definition in state`

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+              return color;
+            }
+          } else {
+            const color = config.default_color;
+            return color;
           }
-          color = state.state === 'on' ? color : config.color_off;
+        } else {
+          const color = colorValue;
+          return color;
```
iantrich: why set `color` to just return it right away? Either return it once at the end or remove the const all together

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+      let iconValue = null;
+      if (configState && configState.icon) {
+        iconValue = configState.icon;
+      } else {
+        iconValue = config.icon;
+      }
+      if (iconValue == 'attribute') {
         if (state) {
           const icon = state.attributes.icon;
           return icon;
```
iantrich:
```suggestion
          return state.attributes.icon;
```

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+          return undefined;
         }
-        return iconOff;
-      }
-      let configState = this.testConfigState(state, config);
-      if (configState && configState.icon) {
-        const icon = configState.icon;
+      } else {
+        const icon = iconValue;
         return icon;
```
iantrich:
```suggestion
        return iconValue;
```

button-card.js
```diff
@@ -121,118 +127,173 @@
...
       }
-      return iconOff;
     }
 
-    blankCardColoredHtml(state, config) {
+    buildStyle(state, config, configState) {
+      let cardStyle = '';
+      let styleArray = undefined;
+      if (state) {
+        if (configState && configState.style)
```
iantrich: This code could use some linting love.
```suggestion
        if (configState && configState.style) {
            styleArray = configState.style;
        }
```

I'm not going to be a huge stickler or anything, but for the sake of my eyes I'll throw out some suggestions :)

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+          styleArray = configState.style;
+        else if (config.style)
```
iantrich: add braces

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+          styleArray = config.style;
+      } else {
+        if (config.style)
```
iantrich: add braces

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+      if (!config.tap_action && state) {
+        switch (state.entity_id.split('.', 2)[0]) {
+          case 'sensor':
+          case 'binary_sensor':
+            return false
+          default:
+            return true
+        }
+      } else {
+        if (config.tap_action && config.tap_action.action == 'none') return false
```
iantrich: add braces
```suggestion
        if (config.tap_action &&
            config.tap_action.action === 'none') {
            return false
        }
```

button-card.js
```diff
@@ -121,118 +127,173 @@
...
+        else return true
```
iantrich: add braces

button-card.js
iantrich: I would personally prefer a `color` variable defined at the top and returned at the bottom

button-card.js
iantrich: Like the color, I think I would prefer a single `icon` variable that is returned at the end of the function

button-card.js
iantrich: Shouldn't the same sensor/binary_sensor check be here if the action is `toggle`?

button-card.js
```diff
@@ -121,118 +127,173 @@
...
-          const value = cssObject[attribute];
-          cardStyle += `${attribute}: ${value};\n`;
-        });
-      }
       this.config.color_type = config.color_type ? config.color_type : 'icon';
-      this.config.color_off = config.color_off ? config.color_off : 'var(--disabled-text-color)';
-      this.config.default_color = config.default_color ? config.default_color : 'var(--primary-text-color)';
-      this.config.card_style = cardStyle;
+      this.config.color_off = 'var(--disabled-text-color)';
+      this.config.default_color = 'var(--primary-text-color)';
```
iantrich: 👍

button-card.js
```diff
@@ -243,33 +304,64 @@
...
-          const event = new Event('hass-more-info', {
-            bubbles: options.bubbles === undefined ? true : options.bubbles,
-            cancelable: Boolean(options.cancelable),
-            composed: options.composed === undefined ? true : options.composed,
-          });
-          event.detail = detail;
-          node.dispatchEvent(event);
-          return event;
+      if (!config.tap_action)
+        config.tap_action = 'toggle'
```
iantrich: this is wrong. you want to set the `action` of `tap_action`
you should also not be mutating the passed argument.
instead we should set config to have a default tap_action if one is not defined in `setConfig`.
I'll add a comment there

button-card.js
```diff
@@ -121,118 +127,173 @@
...
       </ha-card>
-      `;
+        `;
     }
 
     setConfig(config) {
       // if (!config.entity) {
       //   throw new Error('You need to define entity');
       // }
       this.config = { ...config };
```
iantrich: This should allow us to clean up `setConfig` and set the default tap_action's action. 

```suggestion
      this.config = {
          tap_action: {action: "toggle"},
          size: '40%',
          color_type: 'icon',
          color_off: 'var(--disabled-text-color)'
          default_color: 'var(--primary-text-color)'
          ...config
     };
```

button-card.js
```diff
@@ -243,33 +304,64 @@
     }
 
     _toggle(state, config) {
```
iantrich: rename to something like `_handleTap`

button-card.js
```diff
@@ -243,33 +304,64 @@
...
-            bubbles: options.bubbles === undefined ? true : options.bubbles,
-            cancelable: Boolean(options.cancelable),
-            composed: options.composed === undefined ? true : options.composed,
-          });
-          event.detail = detail;
-          node.dispatchEvent(event);
-          return event;
+      if (!config.tap_action)
+        config.tap_action = 'toggle'
+      if (config.tap_action) {
```
iantrich: this check shouldn't be necessary once we have defaults in setConfig

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+        switch (config.tap_action.action) {
+          case 'none':
+            break;
+          case 'more-info':
+          case 'more_info': {
```
iantrich:
```suggestion
          case 'more_info':
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            const detail = { entityId: state.entity_id };
+            const event = new Event('hass-more-info', {
+              bubbles: options.bubbles === undefined ? true : options.bubbles,
+              cancelable: Boolean(options.cancelable),
+              composed: options.composed === undefined ? true : options.composed,
+            });
+            event.detail = detail;
+            node.dispatchEvent(event);
+            return event;
+          }
```
iantrich:
```suggestion

```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            const options = {};
+            const detail = { entityId: state.entity_id };
+            const event = new Event('hass-more-info', {
+              bubbles: options.bubbles === undefined ? true : options.bubbles,
+              cancelable: Boolean(options.cancelable),
+              composed: options.composed === undefined ? true : options.composed,
+            });
+            event.detail = detail;
+            node.dispatchEvent(event);
+            return event;
```
iantrich:
```suggestion
            break;
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+          }
+          case 'location': {
```
iantrich:
```suggestion
          case 'location':
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            root = root && root.shadowRoot || root;
+            root = root && root.querySelector("ha-panel-lovelace");
+            root = root && root.shadowRoot;
+            root = root && root.querySelector("hui-root");
+            root = root && root.shadowRoot;
+            root = root && root.querySelector("ha-app-layout #view");
+            root = root && root.firstElementChild;
+            if (root) root.dispatchEvent(event);
+            break;
+          }
```
iantrich:
```suggestion

```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+          case 'service':
+          case 'call-service': {
```
iantrich:
```suggestion
          case 'call-service':
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            const [domain, service] = config.tap_action.service.split('.', 2);
+            this.hass.callService(domain, service, config.tap_action.service_data);
+            break;
+          }
```
iantrich:
```suggestion

```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            const node = this.shadowRoot;
+            const options = {};
+            const detail = { entityId: state.entity_id };
+            const event = new Event('hass-more-info', {
+              bubbles: options.bubbles === undefined ? true : options.bubbles,
+              cancelable: Boolean(options.cancelable),
+              composed: options.composed === undefined ? true : options.composed,
+            });
+            event.detail = detail;
+            node.dispatchEvent(event);
```
iantrich:
```suggestion
            this.shadowRoot.dispatchEvent(event);
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
-          return event;
+      if (!config.tap_action)
+        config.tap_action = 'toggle'
+      if (config.tap_action) {
+        switch (config.tap_action.action) {
+          case 'none':
+            break;
+          case 'more-info':
+          case 'more_info': {
+            const node = this.shadowRoot;
```
iantrich:
```suggestion

```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            const options = {};
+            const detail = { entityId: state.entity_id };
+            const event = new Event('hass-more-info', {
+              bubbles: options.bubbles === undefined ? true : options.bubbles,
+              cancelable: Boolean(options.cancelable),
+              composed: options.composed === undefined ? true : options.composed,
+            });
+            event.detail = detail;
```
iantrich:
```suggestion
            event.detail = { entityId: state.entity_id };
```

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+        config.tap_action = 'toggle'
+      if (config.tap_action) {
+        switch (config.tap_action.action) {
+          case 'none':
+            break;
+          case 'more-info':
+          case 'more_info': {
+            const node = this.shadowRoot;
+            const options = {};
+            const detail = { entityId: state.entity_id };
```
iantrich:
```suggestion

```

button-card.js
iantrich: Let's rename this to `navigate` to align with HA core

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            root = root && root.shadowRoot;
+            root = root && root.querySelector("app-drawer-layout partial-panel-resolver");
+            root = root && root.shadowRoot || root;
+            root = root && root.querySelector("ha-panel-lovelace");
+            root = root && root.shadowRoot;
+            root = root && root.querySelector("hui-root");
+            root = root && root.shadowRoot;
+            root = root && root.querySelector("ha-app-layout #view");
+            root = root && root.firstElementChild;
+            if (root) root.dispatchEvent(event);
```
iantrich: Why the extract checks here for shadowRoot as opposed to `more-info`?

Might want to take a look at my code for the [radial-menu](https://github.com/custom-cards/radial-menu/blob/master/dist/radial-menu.js#L2546)

Hopefully we can get HA core to expose these functions at some point in the future so each card isn't rewriting this stuff :)

The main take away, is that node is just `this`

button-card.js
```diff
@@ -243,33 +304,64 @@
...
+            break;
+          }
+          case 'service':
+          case 'call-service': {
+            const [domain, service] = config.tap_action.service.split('.', 2);
```
iantrich:
```suggestion
            if (!config.tap_action.service) {
                return;
            }
            const [domain, service] = config.tap_action.service.split('.', 2);
```

RomRider:
> Take a look at my comments and let me know if there is anything you disagree with.

I think I covered all the things on my latest commit :)

button-card.js
RomRider: I changed the way it works a bit, not needed anymore

button-card.js
RomRider: I'm kind of learning so I checked some other code on how it was done and copied it here 😊 
But it's now fixed

button-card.js
```diff
@@ -303,58 +306,41 @@
...
-          case 'more_info': {
-            const node = this.shadowRoot;
-            const options = {};
-            const detail = { entityId: state.entity_id };
-            const event = new Event('hass-more-info', {
-              bubbles: options.bubbles === undefined ? true : options.bubbles,
-              cancelable: Boolean(options.cancelable),
-              composed: options.composed === undefined ? true : options.composed,
+          case 'more_info':
+            event = new Event('hass-more-info', {
```
iantrich: `event` is not needed outside the individual cases. Leave it as a `const` for each case that needs it

RomRider: I changed it like this because without the braces on the `case` block, the scope of variables are at the `switch` level and not at each `case` level so the browser was complaining about `event` being already defined...

iantrich:
Can we cut-off new features in this PR? Each new feature after my initial review makes it more difficult for me to finish a review.

RomRider:
Okay, will do, thanks

button-card.js
```diff
@@ -242,34 +443,48 @@
...
-          node.dispatchEvent(event);
-          return event;
+    _handleTap(state, config) {
+      if (config.tap_action) {
+        let event;
+        switch (config.tap_action.action) {
+          case 'none':
+            break;
+          case 'more-info':
+          case 'more_info':
```
iantrich: I think we should just deprecate `more_info` now. I'm going to merge this request and then do that real quick in a follow-up

Issue #90, opened 2019-04-19
FR: Spin Animation


SeanPM5:
I've long wanted to be able to make certain icons spin, mainly `mdi:fan` to mimic how fan looks both in real life and also how Apple's HomeKit app does it - [4:34 in this video for an example](https://youtu.be/ZRiDWIHKuz8?t=274). 

When @RomRider added the nice blink animation to this card yesterday it reminded me of this item on my wishlist and I tried to see if I could add a spin animation too.

In line 158 of `button-card.js` I added the following:

```
        @keyframes spin {
          from {transform:rotate(0deg);}
          to {transform:rotate(360deg);}
        }
```

And then for my button I was using this code (last line is the important one).
```
          - type: "custom:button-card"
            entity: switch.fan
            # name: Fan
            icon: mdi:fan
            color_type: card
            tap_action:
              action: toggle
            state:
              - value: 'on'
                color: gray #"#ffc845"
                icon: mdi:fan
                style:
                  - animation: spin .7s linear infinite
```

Here is how it looked: https://gfycat.com/BarePoisedConey

Obviously it needs some further refinement, the state is not supposed to spin alongside the icon for starters. I know absolutely nothing about CSS or javascript though, it took me several hours to even get this far.  So I am hoping someone who also likes this idea can take it over the finish line and add this feature perhaps?

Thanks.

PR #91, opened 2019-04-19
Add spin option to state object

iantrich:
Fixes https://github.com/custom-cards/button-card/issues/90


RomRider:
@iantrich, should we also add a`spin` option for the main config?
Up until now most config options are duplicated in main config and state config which makes sense.

iantrich:
So just an always spinning icon?

RomRider:
Yes, that was my idea, to be able to something like:
For a label card
```yaml
- type: "custom:button-card"
  color_type: label-card
  name: Fans
  icon: mdi:fan
  spin: true
```

For an entity:
```yaml
- type: "custom:button-card"
  entity: switch.fan
  icon: mdi:link
  state:
    - value: 'on'
      spin: true
    - value: 'off'
      spin: false

iantrich:
I don't think we should add it to the main. If someone wants it to always spin they can define all states.

RomRider:
> If someone wants it to always spin they can define all states.
There's no state for the `label-card` for example as there's no entity defined

iantrich:
Let's not try to cover every and all edge case and make this more complex than it already is.

PR #95, opened 2019-04-19
Name per state and new show_name parameter

RomRider:
Possibility to define a `name` per state and defaults to retrieving entity's name if no `name` is provided

**Breaking change**
* Introduces a new `show_name` boolean parameter (default is `true`) to display the name or not. (Before if the `name` property was ommited, nothing would be displayed. Now it uses entity`s friendly name or entity's name by default.


README.md
```diff
@@ -38,21 +38,22 @@ Lovelace Button card for your entities.
...
+| `state`        | object list | optional     | See [State](#State)                              | State to use for the color, icon and style of the button. Multiple states can be defined                                                                                                                                                                                                                                                      |
+| `confirmation` | boolean     | `false`      | `true` \| `false`                                | Show a confirmation popup on tap                                                                                                                                                                                                                                                                                                              |
```
iantrich: You don't need to worry about this spacing unless it makes it easier for you to edit. I just bring up the preview pane in VS Code.

iantrich:
Fix the conflicts and this is good to go

README.md
RomRider: Yeah, the plug-in I use does that automatically but it's a mess to merge on conflicts and not very clear about what really changed, I'm going to disable this feature :)

PR #97, opened 2019-04-19
Get rid of the custom element

### LINKED ISSUE #89: Icon allignment and button with
Playing with the custom:mini-media-player, and adding speaker overview to see which speakers are are used by the Rotels input sources.

Ten button cards in a horizontal-stack, using color_type: blank-card, for empty space.

Looks like when decreasing the button, the button icon does not stay in the middle of the button.
Just nit-picking, any way to make the icon stay in the center?

Any reason its not possible to declare the width of a button (now it auto adjust depending on the card and number of collums)?
4A425D51-D2A2-491F-8D31-0175065E7C4A

### LINKED ISSUE #93: Button stays selected after click and highlight on hover
See:
https://community.home-assistant.io/t/lovelace-button-card/65981/583

RomRider:
Cleanup the code to delete the unneeded custom element. Replaced by a bunch of div. This fixes most if not all the issues introduced by the `mwc-button`.

* It fixes the button staying selected after a click (Fix #93)
Apr-19-2019 13-22-21

* It fixes misalignment issues (Fix #89)
Home_Assistant

* Introduce the same ripple effect as the default button

Everything else stays exactly the same! :)

### DIFF
--- a/button-card.js
+++ b/button-card.js
@@ -102,25 +102,9 @@ export default function domainIcon(domain, state) {
   }
 }
 
-((LitElement, ButtonBase) => {
+((LitElement) => {
   var html = LitElement.prototype.html;
   var css = LitElement.prototype.css;
-
-  customElements.define(
-    'button-card-button',
-    class extends ButtonBase {
-      static get styles() {
-        return css`
-          ${ButtonBase.styles}
-          .mdc-button {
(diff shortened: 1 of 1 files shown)


iantrich:
Looks like this needs to be rebased as well

PR #122, opened 2019-04-24
typescript refactor

### LINKED ISSUE #110: Toggle doesn't work for locks
I believe this is a HA limitation, but toggle action doesn't toggle lock domain entities.  My simple workaround is as follows:

Change lines 542-547:
```
          case 'toggle':
          default:
            this.hass.callService('homeassistant', 'toggle', {
              entity_id: state.entity_id,
            });
            break;
```
to
```
          case 'toggle':
          default:
            if( state.entity_id.split('.', 2)[0] == 'lock'){
              if( state.state == 'locked' ){
                this.hass.callService('lock', 'unlock', {
                  entity_id: state.entity_id,
                });
              } else if( state.state == 'unlocked' ){
                this.hass.callService('lock', 'lock', {
                  entity_id: state.entity_id,
                });
              }
            } else if( state.entity_id.split('.', 2)[0] == 'cover'){
              if( state.state == 'open' ){
                this.hass.callService('cover', 'close_cover', {
                  entity_id: state.entity_id,
                });
              } else if( state.state == 'closed' ){
                this.hass.callService('cover', 'open_cover', {
                  entity_id: state.entity_id,
                });
              }
            } else {
              this.hass.callService('homeassistant', 'toggle', {
                entity_id: state.entity_id,
              });
            }
            break;
```

Hopefully this helps others. 

EDIT: Apparently this is also true for covers - added cover functionality too.

RomRider:
**Features**
* Complete typescript refactor

**Fixes**
* Toggle works now for locks and covers: Fix #110 
* Fix Long Press support in general
* Fix Background color calculation in certain case


RomRider:
We need to update the documentation as well as the support for custom_updater, as we shouldn't include the build in the repo but it should only be in the release.

iantrich:
Custom updater should be fine

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+        return (elt.value == state.state)
+      }
+    })
+    if (!retval && def) {
+      return def;
+    }
+    return retval;
+  }
+
+  private getDefaultColorForState(state: HassEntity) {
```
iantrich:
```suggestion
  private getDefaultColorForState(state: HassEntity): string {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    } else {
+      if (state) {
+        name = state.attributes && state.attributes.friendly_name ?
+          state.attributes.friendly_name : helpers.computeEntity(state.entity_id);
+      }
+    }
+    return name;
+  }
+
+  private buildStateString(state: HassEntity | null) {
```
iantrich:
```suggestion
  private buildStateString(state: HassEntity | undefined): string | undefined {
```
Always use `undefined` and not `null` per the [TypeScript spec](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    let stateString: string | null = null;
```
iantrich:
```suggestion
    let stateString: string | undefined = undefined ;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      styleArray.forEach((cssObject) => {
+        const attribute = Object.keys(cssObject)[0];
+        const value = cssObject[attribute];
+        cardStyle += `${attribute}: ${value}; `;
+      });
+    }
+    return cardStyle;
+  }
+
+  private buildName(state: HassEntity | null, configState: StateConfig | undefined) {
```
iantrich:
```suggestion
  private buildName(state: HassEntity | undefined, configState: StateConfig | undefined): string | undefined {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    if (this.config!.show_name === false) {
+      return null;
```
iantrich:
```suggestion
      return undefined ;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    }
+    let name: string | null = null;
```
iantrich:
```suggestion
    let name: string | undefined = undefined ;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      if (state && state.attributes) {
+        icon = state.attributes.icon ?
+          state.attributes.icon :
+          domainIcon(helpers.computeDomain(state.entity_id), state.state);
+      }
+    }
+    return icon;
+  }
+
+  private buildStyle(state: HassEntity | null, configState: StateConfig | undefined) {
```
iantrich:
```suggestion
  private buildStyle(state: HassEntity | undefined , configState: StateConfig | undefined): string {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+          color = this.config!.default_color;
+        }
+      } else {
+        color = colorValue;
+      }
+    }
+    return color;
+  }
+
+  private buildIcon(state: HassEntity | null, configState: StateConfig | undefined) {
```
iantrich:
```suggestion
  private buildIcon(state: HassEntity | undefined, configState: StateConfig | undefined): string | undefined {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    let icon: undefined | string = undefined;
+    if (configState && configState.icon) {
+      icon = configState.icon;
+    } else if (this.config!.icon) {
+      icon = this.config!.icon;
+    } else {
```
iantrich: put this into a single `else if` rather than an `if` inside an `else`

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+          -moz-animation: rotating 2s linear infinite;
+          -ms-animation: rotating 2s linear infinite;
+          -o-animation: rotating 2s linear infinite;
+          animation: rotating 2s linear infinite;
+        }
+      `;
+  }
+
+  protected render(): TemplateResult | void {
+    let state: HassEntity | null = null;
```
iantrich:
```suggestion
    let state: HassEntity | undefined = undefined;
```

iantrich: Add a guard
```suggestion
   if (!this.config || !this.hass) {
     return html``;
   }
    let state: HassEntity | null = null;
```

iantrich: Actually we can do better and with the guard below, this gets cleaner as the linter won't complain about config or hass being undefined

```suggestion
    const state = this.config.entity ? this.hass.states[this.config.entity] : undefined;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    if (this.config!.entity) {
+      state = this.hass!.states[this.config!.entity];
+    }
+    const configState = this.testConfigState(state)
+    if (this.config)
```
iantrich: Remove this as we are checking this earlier to avoid unnecessary work

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    if (colorObj.isValid && colorObj.getLuminance() > 0.5) {
+      fontColor = 'rgb(62, 62, 62)'; // bright colors - black font
+    } else {
+      fontColor = 'rgb(234, 234, 234)';// dark colors - white font
+    }
+    return fontColor;
+  }
+
+  private testConfigState(state: HassEntity | null): undefined | StateConfig {
+    if (!state || !this.config || !this.config.state) {
```
iantrich:
```suggestion
    if (!state || !this.config.state) {
```
We're already testing this in `render`

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      return undefined;
+    }
+    var retval: StateConfig | undefined = undefined;
```
iantrich: don't use `var`

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      }
+    })
+    if (!retval && def) {
+      return def;
+    }
+    return retval;
+  }
+
+  private getDefaultColorForState(state: HassEntity) {
+    if (this.config) {
```
iantrich: don't need to check as we have it in `render`

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    return '';
+  }
+
+  private buildCssColorAttribute(state: HassEntity | null, configState: StateConfig | undefined) {
+    let colorValue: string = '';
+    let color: undefined | string = undefined;
+    if (configState && configState.color) {
+      colorValue = configState.color;
+    } else {
+
```
iantrich: remove blank line

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      if (units) {
+        stateString = state.state + " " + units;
+      } else {
+        stateString = state.state
+      }
+    }
+    return stateString;
+  }
+
+  private buildUnits(state: HassEntity | null) {
```
iantrich:
```suggestion
  private buildUnits(state: HassEntity | undefined): string | undefined {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    let units: string | null = null;
```
iantrich:
```suggestion
    let units: string | undefined= undefined;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+          units = state.attributes.unit_of_measurement;
+        } else {
+          units = this.config!.units ? this.config!.units : null;
+        }
+      }
+    }
+    return units;
+  }
+
+  private buildNameStateConcat(name: string | null, stateString: string | null) {
```
iantrich:
```suggestion
  private buildNameStateConcat(name: string | undefined, stateString: string | undefined): string | undefined {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    if (!name && !stateString) {
+      return null;
```
iantrich:
```suggestion
      return undefined;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    }
+    let nameStateString: string | null = null;
```
iantrich:
```suggestion
    let nameStateString: string | undefined= undefined;
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+    if (stateString !== null) {
```
iantrich:
```suggestion
    if (stateString) {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      } else {
+        nameStateString = stateString;
+      }
+    } else {
+      nameStateString = name;
+    }
+    return nameStateString;
+  }
+
+  private isClickable(state: HassEntity | null): boolean {
```
iantrich:
```suggestion
  private isClickable(state: HassEntity | undefined): boolean {
```

src/button-card.ts
```diff
@@ -0,0 +1,569 @@
...
+      }
+    }
+    return clickable;
+  }
+
+  private rotate(configState: StateConfig | undefined): string {
+    return configState && configState.spin ? 'rotating' : '';
+  }
+
+  private buttonContent(state: HassEntity | null, configState: StateConfig | undefined, color: string): TemplateResult {
```
iantrich:
```suggestion
  private buttonContent(state: HassEntity | undefined, configState: StateConfig | undefined, color: string): TemplateResult {
```

src/button-card.ts
```diff
@@ -0,0 +1,558 @@
...
+  private _handleTap(ev): void {
+    if (this.config!.confirmation &&
+      !confirm(this.config!.confirmation)) {
+      return;
+    }
+    const config = ev.target.config;
+    handleClick(this, this.hass!, config, false);
+  }
+
+  private _handleHold(ev): void {
```
iantrich:
```suggestion
  private _handleHold(): void {
```
You're not using the event so don't have it as an input. Lit is smart about not passing it if you don't need it

src/button-card.ts
```diff
@@ -0,0 +1,558 @@
...
+    this.config!.color_on = 'var(--paper-item-icon-active-color)';
+  }
+
+  // The height of your card. Home Assistant uses this to automatically
+  // distribute all cards over the available columns.
+  public getCardSize(): number {
+    return 3;
+  }
+
+  private _handleTap(ev): void {
```
iantrich:
```suggestion
  private _handleTap(): void {
```

src/button-card.ts
iantrich: Oh wait, you are using it. Sorry

src/button-card.ts
```diff
@@ -0,0 +1,558 @@
...
+      hold_action: { action: "none" },
+      size: '40%',
+      color_type: 'icon',
+      show_name: true,
+      show_state: false,
+      show_icon: true,
+      show_units: true,
+      ...config
+    };
+    this.config!.default_color = 'var(--primary-text-color)';
```
iantrich: Why put these outside the initial definition?

README.md
```diff
@@ -36,26 +36,20 @@ Lovelace Button card for your entities.
 - confirmation popup for sensitive items (optional)
 - support for [custom_updater](https://github.com/custom-components/custom_updater)
 
-## :warning: Long Press Support :warning:
```
iantrich: This warning is still valid. You'd have to package mwc-ripple with the card to get around this

package.json
```diff
@@ -1,27 +1,50 @@
...
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/kuuji/button-card.git"
+    "url": "git+https://github.com/custom-cards/button-card.git"
   },
   "keywords": [
     "lovelace"
   ],
   "author": "kuuji",
```
iantrich: add `contributors`
https://docs.npmjs.com/files/package.json#people-fields-author-contributors

src/button-card.ts
```diff
@@ -0,0 +1,558 @@
...
+          -webkit-animation: rotating 2s linear infinite;
+          -moz-animation: rotating 2s linear infinite;
+          -ms-animation: rotating 2s linear infinite;
+          -o-animation: rotating 2s linear infinite;
+          animation: rotating 2s linear infinite;
+        }
+      `;
+  }
+
+  protected render(): TemplateResult | void {
```
iantrich: We should probably add a `shouldUpdate`. Right now, any time there is a `hass` event, Lit is going to see if we need to update the button, creating unnecessary work. For us we need to check if the config has changed or the state/attributes of the entity if defined

https://github.com/home-assistant/home-assistant-polymer/blob/dev/src/panels/lovelace/cards/hui-picture-entity-card.ts#L54

src/button-card.ts
RomRider: Because that's what I wanted. It's not an option to configure those 😊
Maybe I should define another property for those specifically?

README.md
RomRider: I'm not sure I understand why `mwc-ripple` would be linked to long press?
I deleted the reference to the import `paper-ripple`, but kept the element until I figure out how to use it without errors.
`mwc-ripple` is just for the ripple effect as far as I know.

src/button-card.ts
iantrich: No. They can use a theme

README.md
iantrich: You're right, I believe :+1:

PR #126, opened 2019-04-26
Correct mistake in examples

coxtor:
changed "service" to "call-service", realizing this error took longer than it should have :)


iantrich:
Thanks!

PR #132, opened 2019-04-28
Switch to css grid and code refactor

### LINKED ISSUE #15: FR: scale icon while leaving button same size
The description for size says it resizes the icon, but it actually seems to resize both the icon and the button. 

Nothing wrong with that (although you might consider updating the documentation)... except I'd like to scale just the icon while leaving the overall button size unchanged (some icons look too small/large next to others).  It's a minor issue, so if it's more trouble than it's worth, no biggie.  Only challenge might be how to handle attempts to scale the icon larger than the button - maybe there is a limit of +/- 25% or something.

Thank you very much for this card - the ability to change the color of the whole button on action is especially nice!

Unrelated: your examples seems to imply you have a dark theme for lovelace?  Could you share how you're doing that?

### LINKED ISSUE #29: Style: width no longer recognized
Before the latest HA update, I could set the css style 'width: 675px' to modify the width of a column or vertical stack. Since the last HA update, it no longer works, or recognizes that setting.

### LINKED ISSUE #79: Text only buttons became too small to press on latest release using v90.2 hass
I have a bunch of buttons defined as below which used to create a quite useful sized button. They don't have icons because there aren't really any I want there. As of the latest update running on v90.2, they have gotten really really short, just barely enough to hold the text. Is there some way to make them taller again (ie, add a border above and below the text)?

```
        - type: "custom:button-card"
          name: Jazz
          action: service
          service:
            domain: media_player
            action: play_media 
            data:
              entity_id: "media_player.itunes"
              media_content_id: "best-jazz"
              media_content_type: "playlist"
```

### LINKED ISSUE #111: Feature Request: button height
There are times, the button doesn’t line up with other components when in a horizontal stack. It would be nice to manually set the height of the button in these cases. Also if using a few Buttons together in a horizontal stack they are too tall. You can shrink them by using blank buttons but it would be nice to have them horizontally take the full width and just manually reduce the height.

RomRider:
**New Features**
* Complete rewrite of the card using CSS grids for easier maintenance
* `size` and CSS `height` are now working properly. Fixes #15, Fixes #111, Fixes #79 
* Ability to specify the width of the card through styles (card without `width` specified will occupy the remaining space). Fixes #29 
  image

  ```yaml
      - type: horizontal-stack
        cards:
          - type: "custom:button-card"
            entity: light.test_light
            color: auto
            name: s:default h:200px
            style:
              - height: 200px
          - type: "custom:button-card"
            entity: light.test_light
            color_type: card
            color: auto
            name: s:100% h:200px
            size: 100%
            style:
              - height: 200px
          - type: "custom:button-card"
            entity: light.test_light
            color_type: card
            color: auto
            size: 10%
            name: s:10% h:200px
            style:
              - height: 200px
      - type: horizontal-stack
        cards:
          - type: "custom:button-card"
            entity: light.test_light
            color: auto
            name: 60px
            style:
              - height: 60px
              - width: 60px
          - type: "custom:button-card"
            entity: light.test_light
            color_type: card
            color: auto
            name: 80px
            style:
              - height: 80px
              - width: 30px
          - type: "custom:button-card"
            entity: light.test_light
            color_type: card
            color: auto
            name: 300px
            style:
              - height: 300px
  ```


RomRider:
Will fix #79 also.

apop880:
This looks awesome, and I downloaded it to play around with the new functions. However, one thing I noticed is that I used to be able to use text-align: left to align text on a label-card. That doesn't seem to work anymore, not sure if that was intended or not.

RomRider:
It's not on purpose but I can't satisfy everything at the same time unfortunately :disappointed: 
Since it's using css grids now I think that `self-justify: left` should do the trick. But it's not going to work on those 2 layouts: `icon_name_state2nd` and `icon_state_name2nd`. For all the other layouts it might work (maybe the icon will also be aligned left, if that's the case let me know, I can easily fix it)

Unless we introduce styles for each component inside the card, I don't think this will be possible for the 2 above-mentioned layouts. Please open a separate issue and screenshot what you expect with the config so that I can think about something.

RomRider:
> I think that `justify-items: start` should do the trick

Tested and doesn't work, will think about something for the future...

README.md
```diff
@@ -247,7 +247,7 @@ Horizontal stack with :
       color: rgb(223, 255, 97)
       icon: mdi:volume-plus
       tap_action:
-        action: call-service
+        action: service
```
iantrich: This should still be `call-service`

README.md
```diff
@@ -256,7 +256,7 @@ Horizontal stack with :
       color: rgb(223, 255, 97)
       icon: mdi:volume-minus
       tap_action:
-        action: call-service
+        action: service
```
iantrich: +1

README.md
```diff
@@ -331,7 +331,7 @@ If you don't specify any operator, `==` will be used to match the current state
   entity: input_select.cube_mode
   icon: mdi:cube
   tap_action:
-    action: call-service
+    action: service
```
iantrich: +1

src/button-card.ts
```diff
@@ -457,35 +307,113 @@ class ButtonCard extends LitElement {
...
+        ${stateString ? html`<div class="state">${stateString}</div>` : ''}
+      </div>
+    `;
+  }
+
+  private _getIconHtml(
+    state: HassEntity | undefined,
+    configState: StateConfig | undefined,
+    color: string,
+    // style: StyleInfo,
```
iantrich: remove if not used

src/button-card.ts
```diff
@@ -457,35 +307,113 @@ class ButtonCard extends LitElement {
...
+  ): TemplateResult {
+    const icon = this._buildIcon(state, configState);
+    const entityPicture = this._buildEntityPicture(state, configState);
+    const entityPictureStyleFromConfig = this._buildEntityPictureStyle(state, configState);
+
+    const haIconStyle = {
+      color,
+      width: this.config!.size,
+      'min-width': this.config!.size,
+      // 'min-width': style.height ? '100%' : this.config!.size,
```
iantrich: remove if not used

src/button-card.ts
```diff
@@ -457,35 +307,113 @@ class ButtonCard extends LitElement {
...
+    const entityPictureStyle = {
+      ...haIconStyle,
+      ...entityPictureStyleFromConfig,
+    };
+
+    if (icon || entityPicture) {
+      return html`
+        <div class="img-cell">
+          ${icon && !entityPicture ? html`<ha-icon style=${styleMap(haIconStyle)}
+            icon="${icon}" class="icon ${this._rotate(configState)}"></ha-icon>` : ''}
```
iantrich:
```suggestion
            .icon="${icon}" class="icon ${this._rotate(configState)}"></ha-icon>` : ''}
```

apop880:
No worries, totally understand it's a low priority.

One other thing I did find, when changing the height of a label-card, the text stays bottom-aligned. Previously I was able to use vertical-align: center in style but that doesn't work anymore. However, I hacked around this by adding the following style to the script:

```  .container.name {
    grid-template-areas: "n";
    grid-template-columns: 1fr;
    grid-template-rows: 1fr;
  }
```

And then the following to _buttonContent:

```
case 'name':
                return this._gridHtml(state, configState, this.config.layout, color, nameStateString, undefined);
```

Then I included ```layout: name``` in my YAML.

RomRider:
> One other thing I did find, when changing the height of a label-card, the text stays bottom-aligned. Previously I was able to use vertical-align: center in style but that doesn't work anymore

Nice catch! Will correct this, thanks!

README.md
RomRider: Right, that's strange... I don't remember changing that 😅

RomRider:
@apop880, with this new PR #136, you'll be able to put the text wherever you want, left or right ;)

PR #134, opened 2019-04-29
Label and templates support

### LINKED ISSUE #112: Last changed info on button
Hi, I'd love the ability to add the last changed (secondary info) that exists for the entity card onto the button. 

Thanks

### LINKED ISSUE #120: Feature Request: Show Attribute Instead of State
I'd like an option to display an entity's attribute on a button instead of the state. My use case would be for a light, if it's off I'd like to show the state of Off, but if it's on, I'd like to see the current brightness.

RomRider:
**New features**
* New option `label`: can be used for anything, if template is needed, use `label_template`. Both are also available per `state`. Fixes #112, Fixes #120 
* New option `show_label`: default is `false`. To display the `label` or `label_template` field
* New `layout`: `icon_label` to display the label on the right side of the icon

* templating support 🎉  in `label_template` and `state` with the `template` operator.
  * In `label_template`:
    Home_Assistant
    ```yaml
    - type: "custom:button-card"
      color_type: icon
      entity: light.test_light
      label_template: >
        var bri = states['light.test_light'].attributes.brightness;
        return 'Brightness: ' + (bri ? bri : '0') + '%';
      show_label: true
      size: 15%
      style:
        - height: 100px
    - type: "custom:button-card"
      color_type: icon
      entity: light.test_light
      layout: icon_label
      label_template: >
        return 'Other State: ' + states['switch.skylight'].state;
      show_label: true
      show_name: false
      style:
        - height: 100px
    ```
  * `value` with `template` operator
    ```yaml
    - type: "custom:button-card"
      color_type: icon
      entity: switch.skylight
      show_state: true
      show_label: true
      state:
        - operator: template
          value: >
            return states['light.test_light'].attributes
            && (states['light.test_light'].attributes.brightness <= 100)
          icon: mdi:alert
        - operator: default
          icon: mdi:lightbulb
    - type: "custom:button-card"
      color_type: icon
      entity: light.test_light
      show_label: true
      state:
        - operator: template
          value: >
            return states['input_select.light_mode'].state === 'night_mode'
          icon: mdi:weather-night
          label: Night Mode
        - operator: default
          icon: mdi:white-balance-sunny
          label: Day Mode
    ```

**Fixes**
* Blank card was not following the specified width


iantrich:
I am not a fan of this approach. I really don't think each card should be implementing templates separately and the common solution should be used. Just my opinion.

RomRider:
By common solution you mean the config template card?

iantrich:
That is the current solution; perhaps not the best solution as it is pretty brute-force, but regardless, I'm just seeing lots of requests on numerous other custom cards to add such features and it just seems unnecessary work for everyone to do individually.

RomRider:
@iantrich, I agree, I don't plan to add anything more related to templates, this was just an easy way to display the things that you want in the card. Regarding the config template card there's a bug when used with the button card (when the card first displays).
Fails on `customElements.whebDefined: "custom:button-card" is not a valid element name` A lovelace refresh fixes the issue.

jarrah31:
It's much less complicated to use the built-in templates option than using another custom card. I've experimented with config template card and whilst I have it working well, it adds an extra eight lines of code for each button because I need to reference two entities for my particular setup. 

There's also more existing examples within HA that use templates, so finding existing solutions is easier than trying to work out new ways of doing the same with java script.

Thanks again for this excellent update @RomRider

iantrich:
Like I said, config template card is brute-force templates. There could be a more elegant solution made. That doesn't negate the fact that we're setting a precedent for developers having to repeat work to implement templates in their card. IMO easier/maintainable cards that leverage common practices outweighs extra configuration lines. Whether that be wrapper cards that yield things like css styling or templates or npm modules that give a common interface. I would actually lean more towards the modules than the wrapper cards. Somthing similar to card-tools, but hosted so that it can be bundled. ¯\_(ツ)_/¯

jarrah31:
yep, fair point, and I guess each template implementation is slightly different to the way other cards implement templates, which in turns adds its own complexity. It's all swings and roundabouts! :)

PR #136, opened 2019-05-01
Per element style config

RomRider:
**Breaking Changes**
* for now `style` and `entity_picture_style` are still supported, but please make sure you update your configuration so that `style` becomes `card` and `entity_picture_style` becomes `entity_picture`, both under the `styles` option (note the **S** at the end of `styles`). **Next releases will drop the support for `style` and `entity_picture_style`.**
* The styles defined in each state are now **additive** with the styles defined in the main part of the configuration. It means you don't have to define all the CSS entries in each state, only define the ones which change. Define the common ones in the main `styles` config part and define specific ones in each `state` `style` entries.

**New Features**
* Styling per element supported by using the new `styles` object available in the main config and in each `state` entry.
  You can configure the style for:
  * `card`: Styles that are defined here will be applied to the whole card and it's content, unless redefined in elements below.
  * `entity_picture`
  * `icon`
  * `name`
  * `state`
  * `label`
  per-style-config

  ```yaml
  - type: "custom:button-card"
    color_type: icon
    entity: light.test_light
    label_template: >
      var bri = states['light.test_light'].attributes.brightness;
      return 'Brightness: ' + (bri ? bri : '0') + '%';
    show_label: true
    show_state: true
    size: 10%
    styles:
      card:
        - height: 100px
      label:
        - color: gray
        - font-size: 9px
        - justify-self: start
        - padding: 0px 5px
      name:
        - text-transform: uppercase
        - letter-spacing: 0.5em
        - font-familly: cursive
        - justify-self: start
        - padding: 0px 5px
      state:
        - justify-self: start
        - font-size: 10px
        - padding: 0px 5px
    state:
      - value: 'on'
        styles:
          state:
            - color: green
      - value: 'off'
        styles:
          state:
            - color: red
          card:
            - filter: brightness(40%)
  - type: "custom:button-card"
    color_type: icon
    entity: light.test_light
    layout: icon_label
    label_template: >
      return 'Other State: ' + states['switch.skylight'].state;
    show_label: true
    show_name: false
    size: 100%
    styles:
      card:
        - height: 200px
      label:
        - font-weight: bold
        - writing-mode: vertical-rl
        - text-orientation: mixed
    state:
      - value: 'on'
        styles:
          label:
            - color: red
      - value: 'off'
        styles:
          label:
            - color: green
  ```

**Fixes**
* Brightness calculation for lights was different than the default from Home Assistant. This is fixed.
* Own longpress was not used in some cases


RomRider:
Rebase needed after #135

PR #139, opened 2019-05-02
Add a nice locking mechanism

### LINKED ISSUE #141: Icon colour to match bulb colour (warm/cool white)
**Checklist**

**Describe the bug**
I am attempting to make the icon colour match the bulb colour. I have a Philips Hue White Ambiance bulb that can do warm and cool white, and it's this colour temperature I'd like to see reflected in the icon please.

Looking at the GitHub guide, I see that using `color: auto` should "set the color based on the color of a light." if I have `colour_type` set to `icon` right? In my case though, the icon turns green when off and yellow when on (see screenshots below)

**Version of the card**
Version: 1.5.0

**To Reproduce**
```yaml
              - type: "custom:button-card"
                entity: light.study_desk_lamp
                color_type: icon
                color: auto
                show_name: false
                size: 50%
                hold_action:
                  action: more-info
                styles:
                  card:
                    - width: 100px
                    - height: 100px
```
These are the bulb attributes when switched on:
```
min_mireds: 153
max_mireds: 500
brightness: 254
color_temp: 153
friendly_name: Study desk lamp
supported_features: 43
```
**Screenshots**
Card when Off:
image
Card when On:
image
Colour temp currently set to White:
image

**Expected behavior**
I would like the icon colour to turn white or yellow to match the colour temperature please.

 - Browser - Safari
 - Version - 12.1

RomRider:
**New Features**
* You can return HTML in the `label` or `label_template` to achieve things like:
  image
  ```yaml
  label_template: >
    return 'Connection: '
    + (states['switch.connection'].state === 'on'
        ? '<span style="color: #00FF00;">enabled</span>'
        : '<span style="color: #FF0000;">disabled</span>')
    + ' / '
    + (states['binary_sensor.status'].state === 'on' ? 'connected' : 'disconnected')
  ```
* With `lock: true` You can now lock your button (inspired by the great @thomasloven). This will display a normal button with a lock symbol in the corner. Clicking the button will make the lock go away and enable the button to be manoeuvred for five seconds.
  lock
* `show_last_changed`: If true will replace the label and display the last_changed attribute in a nice way.
  image

**Fixes**
* Support for light temperature with `color: auto`. Fixes #141 
* `blank-card` had a background if a height was defined


iantrich:
image

Getting the following with the latest (could have been from an earlier change, been awhile since I tried this card in my production setup) but looks like our definition of html is conflicting with another card's. Will need to investigate a bit more.

PR #151, opened 2019-05-10
Update with repeat documentation

apop880:
Updating readme to include documentation for repeating `hold_action`.


RomRider:
Good catch, I forgot that one! Thanks a lot! 👍

PR #152, opened 2019-05-11
Add rollup terser to decrease bundle size

iantrich:
Lowers distribution from 172KB to 60KB

### DIFF
--- a/dist/button-card.js
+++ b/dist/button-card.js
@@ -1,3763 +1,692 @@
-/*! *****************************************************************************
-Copyright (c) Microsoft Corporation. All rights reserved.
-Licensed under the Apache License, Version 2.0 (the "License"); you may not use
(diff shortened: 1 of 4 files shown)


RomRider:
Good idea, thanks :)
Why is the `package.lock.json` so much modified? Some versions on your commit are older than the one used before. Is this expected?

iantrich:
weird, I didn't really review the lock file, just ran npm install

iantrich:
I think npm is trying to be smart about possible dependency conflicts.

iantrich:
I think this is fine with the package json diff. What do you think?

RomRider:
I'll integrate this to the next PR if you don't mind and close this one, what do you think? I just don't want to deal with merge conflicts 😅

PR #183, opened 2019-06-04
Support for timer entities

### LINKED ISSUE #156: Help with timer
Hello
i have a button. i have an automation assossiated so that when it is turn on, a timer starts
is there anyway to make the change as per the picture?
basiccaly, if the started starts it is showed on a second line of the button
check:
image

RomRider:
**NEW FEATURES**
* Support for `timer` entities. It will now display the time remaining while running or paused and idle while being idle. Fixes #156

### DIFF
--- a/dist/button-card.js
+++ b/dist/button-card.js
@@ -1,7 +1,7 @@
 function t(t, e, n, i) {
   var r,
       s = arguments.length,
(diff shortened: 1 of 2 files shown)


beliy1985:
It does not work again!

RomRider:
Please open an issue with more details than "it doesn't work"... Thanks

PR #219, opened 2019-09-09
Correct mismatch for state default colors

lambtho12:


RomRider:
Will fix the documentation in the next release. Thanks.

PR #222, opened 2019-10-04
Added safari10 mangle option in terser

trimethia:
On older iOS devices, the current card can't be displayed and outputs the following error:
`button-card.js:69.0 SyntaxError: Cannot declare a const variable twice: 't'`

After a quick search I believe this is due to an [ES6 bug in Safari 10](https://bugs.webkit.org/show_bug.cgi?id=171041).

Terser provides a [mangle option to fix this issue](https://github.com/terser/terser#mangle-options). I verified that the card generated after adding this option to terser and running using `npm run build` is displayed correctly on my iOS device.

### DIFF
--- a/dist/button-card.js
+++ b/dist/button-card.js
@@ -66,7 +66,7 @@ function t(t, e, n, i) {
           let e = -1;for (; -1 !== (e = t.data.indexOf(c, e + 1));) this.parts.push({ type: "node", index: -1 }), o++;
         }
       } else r.currentNode = i.pop();
-    }for (const t of n) t.parentNode.removeChild(t);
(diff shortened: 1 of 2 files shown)


iantrich:
I'll try to do some testing on different devices types with this over the weekend.

@trimethia did you just check on your iOS device?

trimethia:
> @trimethia did you just check on your iOS device?

Yes, it displays fine in my device.

iantrich:
I mean, did you try any other device types?

trimethia:
Also tried safari on a mac and seemed fine.

RomRider:
Will be included in next release, I'm closing this one as I've changed how the module is built. Thanks @trimethia!

PR #231, opened 2019-10-19
Bunch of features

### LINKED ISSUE #192: Add an option to configure the Lock time
**Is your feature request related to a problem? Please describe.**

The default hard coded 5 seconds for the lock is too long.

**Describe the solution you'd like**

**Describe alternatives you've considered**
I’d like to be able to set the lock ‘open’ time in the configuration instead of it being hard wired as it is now.

**Additional context**

### LINKED ISSUE #209: Any card inside button-card
This feature request might be too much or even might be a separate card

Button card specially in new version is easily and highly customizable so the idea is to use it not only as button but also as wrapper for another card.

Now we can embed and style any HTML into buttons, if we could use it also to wrap and style other cards into button this could be an UI style.

Is it realistic to add ability to card so we can do something like this?

```
-  type: custom:button-card
   custom_fields:
      temp:
         card:
           type: entities
           entities:
              - sensor.temp

```

### LINKED ISSUE #217: Confirmation linked to actions
Please consider moving the confirmation option under Actions and allow the text to support templating.  This would allow the confirmation message to act like context aware documentation.

I am using tap and double tap actions, however the same confirmation dialog triggers regardless of the type of action invoked.  It would be useful and flexible to selectively allow (or disable) confirmation messages per action type.  If the field supported templating, the user could confirm the change in state (Are you sure you wish to **disable** the...)

### LINKED ISSUE #220: add lock (permanent-not switchable) and differentiate allow-lock
**Is your feature request related to a problem? Please describe.**

Not really a problem, this is an enhancement of the current lock. Which is in fact a lock-allow,and is much used in my setup
I'd love to have a lock(-not-allowed), short lock, or lock-permanent. Several switches should never be allowed to be switched off and currently there is no way in HA/button-card to assure that (other than not linking the entity).

**Describe the solution you'd like**

Since we already have the `lock: true`, this might come close to a breaking change, which is what I would like to prevent. Hence the addition of a `lock-permanent: true` would seem appropriate.

Starting anew, Id propose `lock: true` to be the permanent lock, and `lock_allow: true` the lock we currently use with a blinking lock for 5 seconds.

**Describe alternatives you've considered**

Only other way I think would be of not entering an entity for the button, so the button can't interact at all with that switch. That would have a whole set of undesired consequences regarding templates for label, name, state, so is not a preferred solution.

there's also https://github.com/thomasloven/lovelace-toggle-lock-entity-row but that still doesnt make it unlockable.

**Additional context**

Not much to add, other than maybe differentiate the icon of the lock-permanent to be mdi:lock-alert

thanks for considering.

### LINKED ISSUE #221: "Custom element doesn't exist" errors in older iOS devices
**Checklist**

**Describe the bug**
A red message with the text `Custom element doesn't exist: button-card` is displayed in older iOS devices (tested with iPhone 6s on Safari running iOS 10.3.3) in Home Assistant. The card displays fine with an up-to-date, desktop version of Safari on a mac. 

When the above error happens in my iOS device, Home Assistant outputs the following log:
`button-card.js:69.0 SyntaxError: Cannot declare a const variable twice: 't'`

After a quick search I believe this is due to a Safari ES6 bug. There seems to be a work-around using mangle options and uglify:
https://bugs.webkit.org/show_bug.cgi?id=171041
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/92#issuecomment-324818437

Is there any way to build the card with any workaround so that it works in older iOS devices? 
I'd rather prefer not to update my iOS device at this time. Thanks in advance. 

**Version of the card**
Version: 2.0.5

**Version of Home Assistant**
Version: 0.99.3

**Expected behavior**
Card is displayed.

### LINKED ISSUE #228: template for (lock) confirmation field
**Is your feature request related to a problem? Please describe.**

no problem, just a feature request
**Describe the solution you'd like**

allow for a template to the confirmation field

```
confirmation: >
  [[[return 'Sure to toggle ' + entity.name + '? If not do nothing.']]]
```
**Describe alternatives you've considered**

verbosity...
**Additional context**

please check: https://community.home-assistant.io/t/lovelace-button-card/65981/2525?u=mariusthvdb

thanks for considering!

### LINKED ISSUE #230: allow lock to be set on Hold action
**Is your feature request related to a problem? Please describe.**

Someone accidentally taps a lock and taps it again...which is kind of easy

**Describe the solution you'd like**

For security reason, I would love to be able to 'lock' the hold-action, (and set tap-action to none or more-info) so that another layer of security's builtin.

**Describe alternatives you've considered**

**Additional context**

please see https://community.home-assistant.io/t/lovelace-button-card/65981/2545?u=mariusthvdb

and code I would suggest:

```
  - type: 'custom:button-card'
        template: button_default_title
        lock: true
        confirmation: Sure to toggle?
        tap_action:
          action: none
        hold_action:
          action: call-service
          service: homeassistant.toggle
          service_data:
            entity_id: group.philips_motion_sensor_switches_all
          haptic: success
```

thanks for considering

### LINKED ISSUE #232: Rename dbltap_action to double_tap_action to align with core

RomRider:
# Breaking Changes (& New Features)
* Drop support for `custom_updater`, please upgrade to [HACS](https://github.com/custom-components/hacs)
* URL actions field has been renamed to `url_path` from `url`:
  old:
  ```yaml
  xxx_action:
    action: url
    url: https://www.google.com
  ```
  new:
  ```yaml
  xxx_action:
    action: url
    url_path: https://www.google.com
  ```
* `dbltap_action` has been renamed to `double_tap_action` to match core (Fix #232)
* `confirmation` is now an object and supports templates, and you can specify a different `confirmation` object for each action. You can also specify exempted users. See [confirmation doc](https://github.com/custom-cards/button-card/blob/master/README.md#confirmation) for more info (Fix #228, Fix #217):
  ```yaml
  confirmation: # If defined, this will apply to all the actions
    text: Are you sure?
    exemptions:
      - user: befc8496799848bda1824f2a8111e30a
  tap_action:
    confirmation:
      # Confirmation only for tap_action.
      # If a default confirmation is defined, it will be override it.
      text: '[[[ return `Toggle ${entity.attributes.friendly_name}? ]]]'
      exemptions:
        - user: befc8496799848bda1824f2a8111e30a
  ```
* `lock` is now an object and `unlock_users` is deprecated.
  `lock.enabled` supports templating and you can set the duration of the unlocked state.
  You can define the type of clicks which will unlock the button with `lock.unlock`. Possible values are `tap`, `hold` and `double_tap`. Default is `tap`
  See [lock doc](https://github.com/custom-cards/button-card/blob/master/README.md#lock-object) for more info (Fix #192, Fix #220, Fix #230).
  Old:
  ```yaml
  lock: true
  unlock_users:
    - test
  ```
  New:
  ```yaml
  lock:
    enabled: '[[[ return entity.state === 'on'; ]]]'
    unlock: hold
    duration: 10
    exemptions:
      - username: test
      - user: befc8496799848bda1824f2a8111e30a
  ```

# New Features
* The version of the card is displayed in the javascript console of your browser. Please verify if you're running the latest version before opening a bug report. If not, update and/or clear your cache:
  image
* You can now embed any card in the custom fields 🎉 (Fix #209). Note that to remove the box-shadow for specific cards, use [card-mod](https://github.com/thomasloven/lovelace-card-mod).
  image
  ```yaml
      - type: custom:button-card
        aspect_ratio: 1/1
        custom_fields:
          graph:
            card:
              type: sensor
              entity: sensor.sensor1
              graph: line
              # style is used by card-mod
              style: |
                ha-card {
                  box-shadow: none;
                }
        styles:
          custom_fields:
            graph:
              - filter: opacity(50%)
              - overflow: unset
          card:
            - overflow: unset
          grid:
            - grid-template-areas: '"i" "n" "graph"'
            - grid-template-columns: 1fr
            - grid-template-rows: 1fr min-content min-content

        entity: light.test_light
        hold_action:
          action: more-info
  ```

# Fixes
* Fix support for Safari 10 (Fix #221)


src/button-card.ts
```diff
@@ -44,6 +44,13 @@ import {
 import { styles } from './styles';
 import myComputeStateDisplay from './compute_state_display';
 
+/* eslint no-console: 0 */
+console.info(
+  `%c  BUTTON-CARD  \n%c Version ${BUTTON_CARD_VERSION} `,
+  'color: orange; font-weight: bold; background: black',
+  'color: white; font-weight: bold; background: dimgray',
+);
+
```
iantrich: nice addition!

src/button-card.ts
```diff
@@ -508,6 +525,16 @@ class ButtonCard extends LitElement {
...
       }
     });
+    Object.keys(cards).forEach((key) => {
+      if (cards[key] != undefined) {
+        const customStyle: StyleInfo = {
+          ...this._buildCustomStyleGeneric(state, configState, key),
+          'grid-area': key,
+        };
+        result = html`${result}
+        <div id=${key} class="ellipsis" style=${styleMap(customStyle)}>${createThing(cards[key], this.hass)}</div>`;
```
iantrich: it's happening!!!!!! :tada:

src/create-thing.ts
```diff
@@ -0,0 +1,54 @@
+import { HomeAssistant, fireEvent } from 'custom-card-helpers';
```
iantrich: why not use the one in custom-card-helpers?

src/button-card.ts
RomRider: It's not working great for now because of how events are handled :D Still need to fix some stuff

src/create-thing.ts
RomRider: I think I've changed something in the code to avoid having to use `custom` everywhere:
```js
  if (tag.startsWith('custom:'))
    tag = tag.substr('custom:'.length);
  else
    tag = `hui-${tag}-card`;
```

iantrich: Good call. We can add that to cch

RomRider: Just issued a PR. Not sure my dependencies didn't break anything though...

iantrich: published to npm; 1.2.7

PR #313, opened 2020-04-17
Add grid-template-areas explanation

Petro31:
Adds a quick explanation on the grid-template-areas style attribute.


RomRider:
Thanks @Petro31!

Issue #335, opened 2020-05-31
Add possibility to overrule standard position of lock


gerard33:
**Is your feature request related to a problem? Please describe.**
First of all, thanks for this great card. Using it for around 80% of my Lovelace configuration at the moment.

I have a FR for something that is not really a problem 😉 , but I would like to be able to change the position of the lock on the button card.

**Describe the solution you'd like**
Make it possible to overrule the standard location of the lock.

**Describe alternatives you've considered**
An alternative is changing the layout of my button-card, but I like it the way it currently is 😄

**Additional context**
![button_card_with_lock](https://user-images.githubusercontent.com/11230573/83362634-1a00c700-a393-11ea-97fa-193dc44b684a.png)

RomRider:
That's already possible:
```yaml
styles:
  lock:
    - justify-content: flex-end
    - align-items: flex-end
```
This would put the lock on the bottom-right corner.

gerard33:
Great, couldn't find this in the readme file 🙈.
Thanks for the quick response! Just added this and works like a charme indeed. Closing the FR.

PR #336, opened 2020-05-31
Override location of lock

gerard33:
Add an example how to override the location of the lock as discussed in #335


RomRider:
Thanks

PR #343, opened 2020-06-14
change the card size to 2 per button

basnijholt:
This is more reasonable. A button is about the size of 2 entity rows.

See for example here:
image

Here the columns aren't of equal length because each button is size three, meaning that it's is the same size as the left bottom most card (three entity rows).


basnijholt:
This is how it looks when including the change 
image

RomRider:
Closing in favor of a configurable card size (baa13b129b2c8356d2f1f4547316bd3ab3bf55d9). Thanks.

PR #437, opened 2021-02-21
minor quotes correction

clemalex824:
Edit enclose quote in lock example otherwise return an error


clemalex824:
reported by an HACF user (https://forum.hacf.fr/t/securite-verrouiller-un-button-card/3124)

RomRider:
Thanks!

PR #631, opened 2022-11-03
fix blank-card

edenhaus:
The blank card shows on HA 2022.11.0 and later are border as show below:
image

With this PR this border is gone:
image

I tested this fix successfully for 2022.10.5 to verify that it is backwards compatible. 

Workaround until this PR is merged:
```yaml
type: custom:button-card
color_type: blank-card
styles:
  card:
    - border-style: none
```


RomRider:
Thanks, I've fixed that without seeing your PR first, sorry.

PR #691, opened 2023-05-02
Fix icon colors

marciogranzotto:
Use the default Home Assisntat's domain state colors


marciogranzotto:
@RomRider can you take a look on this?

RomRider:
Fixed in latest beta (same method as native HA). I'm closing this one. Thanks for the PR.

PR #791, opened 2023-10-27
Translation of README into French

McFlyPartages:
I've taken the liberty of translating the README into French with the addition of the link in the main README. (At first for my personal use)

I don't know if you're interested, but since the doc doesn't move much, it might be interesting to put it in several languages.

You must be the most-used card ;)


dcapslock:
C'est bien! Merci beaucoup. 🙏

RomRider:
Thanks for the time and effort you put into this. I'm a bit worried about maintaining 2 versions of the documentation up to date. The English version is already hard to maintain with all the card features.

It is still updated and will still be in the future (even if I haven't been very present lately 😬) and I wouldn't want users to be misguided by a non up-to-date documentation.

dcapslock:
Is there an online system available that would allow for a translation but also note that it was out of date?

RomRider:
Not that I aware of. But I know how it is... Once it lags behind it is almost impossible to update it to match the new stuff.

dcapslock:
Perhaps https://github.com/ephraimduncan/translate-readme. 

@McFlyPartages are you still interested in assisting with the French readme tramslation. Perhaps you can test the translate-readme action and see how it goes with providing for a translation.

PR #808, opened 2023-12-09
Don't trigger action on touchcancel event

Ongion:
When using the Android Navigation Back Gesture (swipe from side of screen), tap actions are still triggered even though Android sent a touchcancel event.

This changes makes it so that a touchcancel event does not trigger actions.


RomRider:
Thanks

PR #823, opened 2024-01-22
Add support for array-result in triggers_update

psi-4ward:
Add the ability that a template could return an array of entities.

Ie:
```yaml
triggers_update: >-
  [[[ 
    const result = [];
    for(let i = 0; i<10; i++) result.push(`lights.light${i}`);
    return result;
  ]]]
```


RomRider:
Good idea, thanks!
I'd add the possibility to do this also:
```yaml
triggers_update:
  - >-
    [[[ 
      const result = [];
      for(let i = 0; i<10; i++) result.push(`lights.light${i}`);
      return result;
    ]]]
  - sensor.other_entity
```

Also update the documentation to reflect this change, please

RomRider:
Closing, `triggers_update` is not required anymore starting from 6.1.0-dev.2

Issue #200, opened 2019-07-24
more-info dialog always opens on tap if button-card is placed inside entities card


hcoohb:
**Checklist**
- [x] I updated the card to the latest version available
- [x] I cleared the cache of my browser

**Describe the bug**
If the button card is placed inside and entities card, no matter what tap_action I provide, the more-info dialog opens (and the action is also performed)

**Version of the card**
Version: 1.11.1

**To Reproduce**
This is the configuration I used:
```yaml

  - type: entities
    entities:
    - type: "custom:button-card"
      entity: switch.sw1
      name: Button in entities
      tap_action: 
        action: toggle
      
  - type: "custom:button-card"
    entity: switch.sw1
    name: Button
    tap_action: 
      action: toggle
```

**Expected behavior**
I would expect the more-dialog to not show unless it is the action provided in tap_action

**Desktop (please complete the following information):**
 - Browser chrome Win, Linux
 - Version 75

**Smartphone (please complete the following information):**
 - Device: oneplus6
 - OS: Android 9
 - Browser Chrome
 - Version ?

**Additional context**
Only seems to appear when the button card is inside entities, however, it is the only way I have found to be able to stack vertically 2 buttons and be able to style the background nicely using card-mod

RomRider:
Button card is not supposed to be used inside an entity card. I can't change that behavior, it's how the entity card works unfortunately.

hcoohb:
Thanks for the quick answer!
Oh, I thought it would have been possible to stop the propagation of the click on the button-card.
I have tried with vertical-stack-in-card but because it is not a ha-card, I cannot modify the style (ie margin, background and borders) of the card with card-mod. (as this vertical-stack-in-card would be inside another vertical-stack-in-card and horizontal-card).

What I am trying to do is to have the icon part to toggle a service and the name part of the card to be a link toward the service url. But it seems if I just do a <a></a> inside the name_template, it goes to the url but it also toggle the service at the same time which I do not want.
Would you have any advice?

Thanks and regards

stefancvetkovic:
I am in same train with you @hcoohb. Always more-info showing up when it in entities type.

iantrich:
[...]

PR #904, opened 2025-01-23
Add stop_propagation config option to stop click or touch events propagation

Zhephyr54:
Useful when using button-card inside another card.
Making it a config option instead of always applying it because of [this comment](https://github.com/custom-cards/button-card/issues/200#issuecomment-516451787).

For an example application, when using a button-card, inside the `title-card` of [lovelace-expander-card](https://github.com/MelleD/lovelace-expander-card) with `title-card-clickable`. In this case, when clicking a button-card with a tap action, letting the click/touch event bubble up would also unintentionally expand/collapse the expander card.

### DIFF
--- a/src/action-handler.ts
+++ b/src/action-handler.ts
@@ -23,6 +23,7 @@ export interface ActionHandlerOptions {
   hasHold?: boolean;
   hasDoubleClick?: boolean;
   disabled?: boolean;
+  stopPropagation?: boolean;
   repeat?: number;
   repeatLimit?: number;
 }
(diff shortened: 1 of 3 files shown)


ngocjohn:
@Zhephyr54 Isn't it a bug in the expander-card? Any card with an action in the title card can trigger an expander... Probably the author of expander doesn't take into account the possibility that someone will use an action other than collapse/expand.

For example, I used the default tile-card for the title card. The result is the same, so this is more a problem of what handler function the expander uses

Zhephyr54:
@ngocjohn Not sure if this can be done in my specific use case: I am using a [stack-in-card](https://github.com/custom-cards/stack-in-card) as a title card, this stack contains several button cards, some without action handlers that should propagate and others with action handlers that should stop propagation. How would you deal with this?

RomRider:
I don't believe this is still relevant.

PR #936, opened 2025-08-15
Readme update - manual installation note for javascript file being blocked.

### LINKED ISSUE #895: Install does not work.
HACS automatic install simply does not work.

fall back to manual

First instruction is to download button-card.js

file does not even exist in the git repository. ( everything is .ts )

dcapslock:
Fixes #895


RomRider:
Thanks!

PR #944, opened 2025-08-19
Fix camera live stream refresh and support aspect ratio & fit mode

### LINKED ISSUE #913: camera slow refresh rate
**Checklist**

**Describe the bug**
I have this config to show a live camera feed on a dashboard..

```
              - show_state: true
                show_name: true
                camera_view: live
                type: custom:button-card
                show_live_stream: true
                entity: camera.outdoorcam
                name: Front Camera
                state_display: >-
                  [[[ if (states['binary_sensor.outdoorcam_car_occupancy'].state
                  == "on") return '&#128663; Present'; else return 'Away'; ]]]
                aspect_ratio: 1/1
                styles:
                  card:
                    - background: rgba(0,0,0,0.3);
                    - border: 0;
                    - color: '#FFFFFF'
                  icon:
                    - width: 100%
```

This *used* to work but recently (weeks?) the camera feed changed to run at 1 frame every 10 seconds, from the browser logs I see a GET request every 10 seconds. Its not pulling the live stream. Pressing the button to open the stream works as expected and the same streams work fine on other dashboards (which use the default HA card).

The camera is being provided by the Frigate integration (but I dont think this is relevant).

**Version of the card**
Version:  4.1.2

 - Firefox 135

 - Device: Fire 7 tablet
 - OS: Android
 - Browser Wallpanel

dcapslock:
Fixes #913 

Note sure how this would have worked as cameraView is a property not attribute. Perhaps in the past lit handled this somehow?

Also include aspectRatio and fitMode that HA has supported for hui-image since 2023.10 (hacs.json updated).

### DIFF
--- a/hacs.json
+++ b/hacs.json
@@ -3,5 +3,5 @@
   "content_in_root": false,
   "render_readme": true,
   "filename": "button-card.js",
-  "homeassistant": "2023.7.0"
+  "homeassistant": "2023.10.0"
 }


RomRider:
Thanks!

Issue #425, opened 2021-01-27
Bug - On press button returns error


Claw256:
**Is your feature request related to a problem? Please describe.**
Yes, I'm trying to use the confirmation option to only show a confirmation message when the entity (MQTT plug) is on, to stop me accidentally turning off my PC when I'm using it in HA. I don't want the confirmation popup to show when the entity if off.

**Describe the solution you'd like**
Add support for enabling or disabling confirmation, with template support, same as the lock option.

**Describe alternatives you've considered**
I tried using the text option for confirmation to show the confirmation text when the entity is on, but that only controls text and not the actual pop up. If there was an option to disable the confirmation popup based on entity state, I would use that.

**Additional context**
What I currently have at the moment:
```
- type: "custom:button-card"
  entity: switch.my_smart_plug
  icon: mdi:power-plug
  color: auto
  confirmation:
    text: >
      [[[
        if (entity.state === "on")
          return `Are you sure you want to turn off ${entity.attributes.friendly_name}?`;
      ]]]
```

If there is another way of doing this, or if I missed something that I should be using, please let me know.

RomRider:
Hey,

You can achieve that already:
```yaml
- type: "custom:button-card"
  entity: switch.my_smart_plug
  icon: mdi:power-plug
  color: auto
  confirmation: |
    [[[
        if (entity.state === "on")
          return { text: `Are you sure you want to turn off ${entity.attributes.friendly_name}?`} ;
    ]]]
```

Claw256:
> Hey,
> 
> You can achieve that already:
> 
> ```yaml
> - type: "custom:button-card"
>   entity: switch.my_smart_plug
>   icon: mdi:power-plug
>   color: auto
>   confirmation: |
>     [[[
>         if (entity.state === "on")
>           return { text: `Are you sure you want to turn off ${entity.attributes.friendly_name}?`} ;
>     ]]]
> ```

That's good to hear. Your specific example there didn't work for me for some reason, as in when I press the card in lovelace the touch animation shows but the plug doesn't turn on or off, it just stays on. Does it work for you for your setup?

RomRider:
Any message in the console of your browser?
Are you sure you are using the latest version? (it's also displayed in the console)

Claw256:
Yes, here is the message I get in the console:
```
BUTTON-CARD   Version 3.4.1


button-card.js:543 Uncaught TypeError: Cannot assign to read only property '0' of string '[[[
    if (entity.state === "on")
[...]

Issue #616, opened 2022-08-18
A tap_action question.


Opontios:
Don't know if it's been already answered, but is it possible to define 2 different actions under tap_action, as 2 different call services? I tried any possible way that I could, but it isn't validated by the card editor. Thanks.

Wombosvideo:
This is what scripts are for. The example below gives you a button to perform actions for when you are not at home, like turning off the lights and the thermostat.

**Card config:**
```yaml
type: custom:button-card
name: Not at home
tap_action:
  action: call-service
  service: script.not_at_home_actions
```

**Script:**
```yaml
alias: Not at home actions
icon: mdi:script
mode: single
sequence:
  - service: light.turn_off
    data: {}
    target:
      entity_id: light.luca
  - service: climate.turn_off
    data: {}
    target:
      entity_id: climate.luca_thermostat
```

Opontios:
I'm aware about the scripts. I wanted to know if it was possible to send 2 different service calls in the card config.

JayChun:
> I'm aware about the scripts. I wanted to know if it was possible to send 2 different service calls in the card config.

I think you need 'multi_calls', please see in my question body. (https://github.com/custom-cards/button-card/issues/620#issue-1387396576)

dcapslock:
Hi,

1. If this is about returning an action object from a template, this is now available in 4.2.0. 
2. If this is about mutiple service calls, a Home Assistant script is the offical support way. You may also be interested in [this](https://github.com/custom-cards/button-card/discussions/971#discussioncomment-14260376).

PR #945, opened 2025-08-19
Allow action to be a template string

### LINKED ISSUE #543: Add Tap-Action to States for conditional actions
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Have the ability to define different tap-actions for each state of the button in the YAML

**Describe alternatives you've considered**
Using Scripts with conditional checks, however I'd like to have conditional confirmation/lockout depending on the state and avoid extra scripts
Using different actions (tap / double tap etc.) for each service call, however it's not as intuitive for the end user
Using a binary toggle, however it doesn't provide monitoring feedback of status

**Additional context**
For a server on/off button. State provided by ping, shutdown by SSH and power on by smart switch / WoL (if it works).
Providing both actions and state monitoring (red/green indicator) on a button is very user intuitive (and cool)

An example configuration YAML could be:

```yaml
type: custom:button-card
icon: mdi:power
entity: binary_sensor.ping_server
show_name: false
state:
  - value: 'on'
    tap_action:
	  action: call-service
	  service: shell_command.server_shutdown
	  confirmation:
	    text: Power down server?
    styles:
      icon:
        - color: green
  - value: 'off'
    tap_action:
	  action: call-service
	  service: (Wol / switch)
    styles:
      icon:
        - color: red
```

Thank you for the amazing mod!

### LINKED ISSUE #611: Copying tap_action into long_tap_action
**Is your feature request related to a problem? Please describe.**
My wife is handicapped and does not have very fine control over her muscles. It is difficult for her to push a UI button, not too long, not too short.

**Describe the solution you'd like**
A simple way to instruct button-card to handle `long_tap_action` the same as a common `tap_action`.

**Describe alternatives you've considered**
Copy (cut/paste) the `tap_action` YAML into the `long_tap_action` for each and every button... Doable but tedious and prone to errors.

### LINKED ISSUE #850: State-dependent actions
I have a button which I would like to call different services on `tap_action` depending on state. 

Ideally the current action format could be implemented in the state list, along with color, icon, styles etc

Current;y I have implemented the following with templates for which service/targets to use:
```
tap_action:
  action: call-service
  service: >-
    [[[ if (variables.boost_on) return "schedule_state.clear_overrides"; else
    return "input_button.press" ]]]
  data: {}
  target:
    entity_id: >-
      [[[ if (variables.boost_on) return "sensor.heating_schedule"; else return
      "input_button.heating_boost" ]]]
```

Which works ok for the moment, but could see getting rather unwieldy with more than 2 states, or more difficult if e.g. wanting to target a label rather than an entity in one of the states

dcapslock:
Closes #850 
Closes #611 
Closes #543 
Provides another solution for #616 
... and maybe some more.

Readme changes describes what is possible with examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Lovelace Button card for your entities.
 
 ### Action
 
-All the fields support templates, see [templates](#javascript-templates).
+All the fields support templates, see [templates](#javascript-templates). You may also provide a single template for the action itself. Such a template needs to return a javascript object that includes the fields below.
 
 | Name | Type | Default | Supported options | Description |
 | --- | --- | --- | --- | --- |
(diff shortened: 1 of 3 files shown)


dcapslock:
Also fixes #425

PR #952, opened 2025-08-23
Support media-source:// for picture_entity

dcapslock:
Closes #937

### DIFF
--- a/src/button-card.ts
+++ b/src/button-card.ts
@@ -3,6 +3,7 @@
 import { LitElement, html, TemplateResult, CSSResult, PropertyValues } from 'lit';
 import { customElement, property, queryAsync, eventOptions } from 'lit/decorators';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { until } from 'lit/directives/until.js';
 import { Ripple } from '@material/mwc-ripple';
 import { RippleHandlers } from '@material/mwc-ripple/ripple-handlers';
 import { styleMap, StyleInfo } from 'lit-html/directives/style-map';
@@ -21,6 +22,7 @@ import {
   ButtonCardEmbeddedCardsConfig,
   ColorType,
   CustomFields,
+  EntityPicture,
 } from './types/types';
 import { actionHandler } from './action-handler';
 import {
@@ -115,8 +117,6 @@ class ButtonCard extends LitElement {
 
   @property() private _timeRemaining?: number;
 
-  @property() private _picture?: string;
-
   @queryAsync('mwc-ripple') private _ripple!: Promise<Ripple | null>;
 
   private _hasTemplate?: boolean;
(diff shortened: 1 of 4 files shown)


RomRider:
Thanks!
What do you think about providing a helper for that too? So that you can use `media-source://` in any place?

dcapslock:
> What do you think about providing a helper for that too? So that you can use `media-source://` in any place?

It would be tricky due to the async nature of the server call. Perhaps we could return a promise but really that would only be usable in expert JavaScript of setting an element async.

Mariusthvdb:
if I may..
it would be nice if we could prevent always having to write the full path to that media share as `media-share://media_source/local/`

I am aware that is what core frontend does, but there too it is unnecessarily complex. it is not without a reason Frontend maps `/config/www/` to `/local/`

maybe button-card could wrap the `media-share://media_source/local/` to something like `/media/` and map that automatically?

example: my image files would be in 
`media-share://media_source/local/images/family/me.jpg`

mapping that would result in

`/media/images/family/me.jpg`

or, if the analogy with the www were to be used:
`/media/local/images/family/me.jpg`

dcapslock:
> if I may.. it would be nice if we could prevent always having to write the full path to that media share as `media-share://media_source/local/`
> 
> I am aware that is what core frontend does, but there too it is unnecessarily complex. it is not without a reason Frontend maps /config/www/ to /local/
> 
> maybe button-card could wrap the `media-share://media_source/local/` to something like /media/ and map that automatically?

I don't agree with this - it may not always be '/local/'. If Frontend comes up with a strategy then we can adopt that. However creating a strategy is prone to be divergent from Frontend and be cuase for confusion if that does happen.

dcapslock:
I recommend that whatever is at the [media_source doco](https://www.home-assistant.io/integrations/media_source/) is what is followed. Users can do their own templating to shorten with variables and configuration templates if they desire.

dcapslock:
@RomRider With Lit and Promises you need to use the until() directive which takes values and promises and renders in priority (passed variables) order. So for providing a helper which returns a promise, all locations where that promise may be used would need to use the until() directive. The PR now update using this approach. Once you have reviewed and considerd the merit of helpers which return a promise, I will leave making a helper for now. Perhaps first a PR that allows for templates to return promises overall is the way to go, and then introduce helpers which return a promise.

Ref: [Lit Asynchronus rendering](https://lit.dev/docs/templates/directives/#asynchronous-rendering)

Mariusthvdb:
> I recommend that whatever is at the [media_source doco](https://www.home-assistant.io/integrations/media_source/) is what is followed. Users can do their own templating to shorten with variables and configuration templates if they desire.

sure.
Agreed to follow what  docs say.
___
just from experience: that documentation is the exact source for confusion, especially the bit under https://www.home-assistant.io/integrations/media_source/#local-media and leaves users in limbo.. ;-)

PR #961, opened 2025-08-28
Stub config section mode

dcapslock:


dcapslock:
Screenshot 2025-08-28 at 1 14 23 pm

Screenshot 2025-08-28 at 1 14 42 pm

src/button-card.ts
```diff
@@ -142,6 +157,27 @@ class ButtonCard extends LitElement {
...
+    const foundEntities = findEntities(hass, maxEntities, entities, entitiesFallback, ['light', 'switch']);
+    if (llCreateCardIsSection) {
+      return {
+        entity: foundEntities[0] || '',
+        section_mode: true,
+        grid_options: {
+          rows: 2,
+          columns: 6,
+          min_rows: 1,
+          min_columns: 1,
```
RomRider, line 171: Those 2 should be removed from the card configuration, it's just useful in getGridOptions.

src/types/types.ts
```diff
@@ -46,6 +46,13 @@ export interface ButtonCardConfig {
   section_mode?: boolean;
 }
 
+export interface GridOptions {
+  rows?: number;
+  columns?: number;
+  min_rows?: number;
+  min_columns?: number;
```
RomRider, line 53: Same as above

RomRider (changes requested):

src/button-card.ts
dcapslock: Will do

RomRider:
Is there any way to get the size selector UI + visibility UI while keeping the yaml editor for the card?

dcapslock:
> Is there any way to get the size selector UI + visibility UI while keeping the yaml editor for the card?

Not cleanly. You can set a congfgElement but but to go to only yaml mode you need to throw a warning in getConfig. Note: visibility works if you set it in yaml.

dcapslock:
@RomRider strategy now update to use Lovelace config for current view.

PR #972, opened 2025-08-30
chore(ci): Add wontfix to label to stale bot

dcapslock:


dcapslock:
See this has already been done.

PR #976, opened 2025-08-30
fix(state_display): Allow empty string or null to override state

### LINKED ISSUE #435: Allow show_state to be templatable
I have a custom button card that I use to show a countdown timer when I press another related button. The countdown only shows properly when you set show_state: true. When the countdown isn't active, I would like to show alternate text without showing the state of the timer (which is "Idle")

It would be great if show_state allowed us to use Javascript templates like many other options. This would allow me to only show the countdown timer when it is active. 

I'm considering using a conditional card, but it would be so nice to do this all in a single card.

image
image

dcapslock:
Fixes #435


RomRider:
Thanks!

PR #987, opened 2025-09-02
fix(color): inactive card background no longer inactive color when `colour_type: card` and `color` set.

### LINKED ISSUE #754: Card color in off state is reverted, when color is defined with RGB
**Checklist**
- [ x ] I updated the card to the latest version available
- [ x ] I cleared the cache of my browser

**Describe the bug**
I have one more issue with card color, if I define color as var(--paper-item-icon-color), then the color of the card in off state is reverted. It should be as the color of covers buttons in screenshot. Same applies for color defined as rgb(0, 122, 255). More info in the previous ticket https://github.com/custom-cards/button-card/issues/737

Light Mode
Screenshot 2023-08-03 at 19 33 07

Dark Mode
Screenshot 2023-08-03 at 19 33 50

```
button_card_templates:
  switch:
    hold_action:
      action: more-info
    tap_action:
      action: toggle
    type: custom:button-card
    color: var(--paper-item-icon-active-color)
    color_type: card
    state:
      - value: 'on'
        styles:
          icon:
            - color: white
          name:
            - color: white
```

dcapslock:
BREAKING CHANGE: Card background color will always be `var (--card-background-color)` when state is inactive and `color_type: card`. You can set card background with state.

Fixes #754


dcapslock:
I hope I have done the commit message correct with BREAKING.

RomRider:
I'm not sure I agree with this change. The idea of the `color_type: card` is to invert the logic.
If color_type is undefined, the logic applies to the icon (inactive is using css variables to match the inactive color of the domain). We should do the same here.

The problem is only when `color_type: card` is set and `color` is set at the same time, in that case, the background doesn't behave correctly it should follow the same logic as the icon if `color_type` is unset, but it keeps using `--card-background-color` instead of using the domain state color. This is what should be fixed, not forcing the background to be `--card-background-color` all the time

RomRider:
Well, actually, if I look at the history, I fixed this in the past and forced the background to be the default card background color when off in this commit https://github.com/custom-cards/button-card/commit/41dea3f72a607ea0fc1c71269bb0122f27970a6e... So it seems it some people already complained about that haha :smile: but it wasn't complete.
So your fix is aligned with that logic

PR #988, opened 2025-09-02
fix(actions): Fix haptic support

### LINKED ISSUE #565: Custom button vibration like a standard button on phones/tablets etc
Hi!

First to say I love this card. Compared to others this is the most customizable thing in the whole Home Assistant so far I saw.

I noticed one mayor difference compared to the standard built in button card in Home Assistant and is that it's not vibrating when I press them on my android phone. Fortunately I added the "Changing the feedback color during a click" option and thats the only thing to let me know I actually clicked the button.

I think it would be cool if it would vibrate too though. Is it feasible?

Thanks!

### LINKED ISSUE #760: Haptic on mobile app/browser doesn't work when combine with "action: navigate"
**Checklist**
- [v ] I updated the card to the latest version available
- [v ] I cleared the cache of my browser

**Describe the bug**
When i set tap_action as "action: navigate", haptic doesn't work, i tried with all haptic options

**Version of the card**
Version: v4.1.1

**To Reproduce**
```yaml
type: custom:button-card
entity: fan.bedroom_fan
tap_action:
  haptic: light
  action: navigate
  navigation_path: /lovelace/fan
hold_action:
  action: toggle
  haptic: light
```

**Screenshots**

**Expected behavior**

 - Device: Samsung Galaxy S23 Ultra
 - OS: Android 13
 - Browser: Chrome, Home Assistant Companion
 - Version: Chrome 116.0.5845.92; Home Assistant 2023.8.2-full

**Additional context**

### LINKED ISSUE #956: [FR]: Support default haptic at configuration level
New FR to track this request specifically.

Refer #121 for more detail.

dcapslock:
Closes #956 
Closes #565 
Fixes #760


dcapslock:
Tested on iOS.

RomRider:
It doesn't seem to do anything different on Android whatever the setting (I get the default vibration) while using the Home Assistant Companion App. Isn't the action firing an event because I always get a vibration even if not set?
Also, it seems there's no way to disable haptics for a specific button. This is quite old and I don't remember how it's supposed to work :upside_down_face:

dcapslock:
Checking this out. Looks like "haptic" event is not being intercepted so the regular Frontend haptic is forwarding as well as the custom (seen in ADB debug messages for emaultor). If you try a custom action which is not toggle you will likely get the correct haptic. I will post back findings.

dcapslock:
Will need a different approach. Will close this PR and submit new to not have a messy commit history.

dcapslock:
Reopening.

dcapslock:
Frontend [PR#27251](https://github.com/home-assistant/frontend/pull/27251) created to allow for capturing events at window level correctly for "haptic" event.

dcapslock:
> Frontend [PR#27251](https://github.com/home-assistant/frontend/pull/27251) created

This has been approved and merged to Frontend dev.

RomRider:
Could we try to keep the "failure" haptic sent by frontend when the action is misconfigured? Should be easy enough to check if frontend sent a failure haptics and letting it go through instead of killing the event and sending ours?

dcapslock:
> Could we try to keep the "failure" haptic sent by frontend when the action is misconfigured? Should be easy enough to check if frontend sent a failure haptics and letting it go through instead of killing the event and sending ours?

Good idea. Should be easy enough.

dcapslock:
Closing in favour of new PR coming once tested.

PR #989, opened 2025-09-03
fix: Embedded light card handle issue

### LINKED ISSUE #427: Embedding light card as custom_field breaks dimmer handle
**Checklist**
- [X] I updated the card to the latest version available
- [X] I cleared the cache of my browser

**Describe the bug**
Embedding a `light` card as `custom_field` breaks interaction with the dimmer handle of the card. The handle doesn't catch the release of the mouse or finger but keeps following the mouse movement. See animation below. On the left a `light` card embedded in the `button-card`, on the right a pure `light` card.

**Version of the card**
Version: 3.4.1

**To Reproduce**
This is the configuration I used (simplified version without templating and styling but showing same bug):
```
type: 'custom:button-card'
entity: light.shelly_leo_dimmer
styles:
  custom_fields:
    setting:
      - position: absolute
      - top: 0%
      - left: 0%
      - width: 100%
      - height: 100%
custom_fields:
  setting:
    card:
      type: light
      icon: 'mdi:arrow-left'
      entity: light.shelly_leo_dimmer
      name: ' '

```

**Screenshots**
see above

**Expected behavior**
The dimmer control should behave the same, no matter if the `light` card is embedded or not.

 - Browser Edge
 - Version 88.0.705.53

not tested

**Additional context**
none

### LINKED ISSUE #901: Problem Honeycomb-Menu after the last update buton-card.
After the last update, when I click on a Honeycomb Menu, the menu doesn't open and all the configuration goes to the left side.
button-card: Version v4.1.2
Honeycomb-Menu: Version 0.14.7
HA : Version 2025.1.2

https://github.com/user-attachments/assets/03072c6a-69fb-4d3d-a4ee-43f0e06afc45

dcapslock:
Fixes #427 
Also will fix #901 though Honeycomb menu should use capture event is their case.

It could be that to prevent any unknown breaking change that a config option is used, but I could not think of a good name for it!

### DIFF
--- a/src/button-card.ts
+++ b/src/button-card.ts
@@ -918,12 +918,12 @@ class ButtonCard extends LitElement {
           <div
             id=${key}
             @action=${this._stopPropagation}
-            @click=${this._stopPropagation}
-            @touchstart=${this._stopPropagation}
-            @mousedown=${this._stopPropagation}
-            @mouseup=${this._stopPropagation}
-            @touchend=${this._stopPropagation}
-            @touchcancel=${this._stopPropagation}
+            @click=${this._sendToDocumet}
+            @touchstart=${this._sendToDocumet}
+            @mousedown=${this._sendToDocumet}
+            @mouseup=${this._sendToDocumet}
+            @touchend=${this._sendToDocumet}
+            @touchcancel=${this._sendToDocumet}
             style=${styleMap(customStyle)}
           >
             ${thing}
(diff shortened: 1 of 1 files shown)


RomRider:
Maybe `propagate_events` default to `true` and some explanation in the README with some use case?

RomRider:
Else could we maybe forward the event to the button-card parent using `this.parentElement` instead of document? Effectively bypassing the card but allowing it to bubble up?

dcapslock:
> `this.parentElement` instead of document

I think this is the best option if you don't think it will be breaking in any way. I can't see that it would be.

RomRider:
It shouldn't break anything but you never know with those events... They give me nightmares 🤣

dcapslock:
Updated.

PR #1005, opened 2025-09-17
fix missing word 'break'

Mariusthvdb:


RomRider:
This is already fixed in the dev branch. Thanks

PR #1010, opened 2025-09-18
fix: Support haptics

dcapslock:
Need to intercept Frontend haptics to be able to use custom. Otherwise custom haptics may not apply.

Best method would have been to intercept the forwardHaptic events. However since they are fired at window, its impossible to capture.

Next best method is to intercept the three type of calls:

1) navigator.vibrate - perhaps used on devices when not using App.
2) Android external app messaging
3) webkit (iOS) external app messaging

See homeassistant.ts types for link to Frontend external messaging code.

Intercept is temporary as we have no idea of what the latest state of the messaging interfaces are. Binding methods is critical espeically with Anroid as these are created by companion app java interface and it will [complain otherwise](https://github.com/react-native-webview/react-native-webview/issues/323#issuecomment-511824940).

### DIFF
--- a/src/haptic.ts
+++ b/src/haptic.ts
@@ -1,6 +1,8 @@
 import { fireEvent } from './common/fire-event';
 import { EMMessage, EMOutgoingMessageHaptic, isExternal } from './types/homeassistant';
 
+// https://github.com/home-assistant/frontend/blob/master/src/data/haptics.ts
+
 // Allowed types are from iOS HIG.
 // https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/feedback/#haptics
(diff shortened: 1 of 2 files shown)


dcapslock:
@RomRider draft for now so you can see this solution. Confirming there is no way that the browser events can be captured as testing shows that as the target is window, the event phase goes straight to target and bubbling listeners are fired before capture. So intercepting function/app calls is next best. I have tested quite robustly, and am happy that this will work, and only gets used if user sets `haptic:` so if the code would for some reason break then users just remove haptic.

dcapslock:
Closing in favour of fixing Frontend so the first simpler option can be used.

PR #360, opened 2020-07-15
Add pressed and released event to allow usage as momentary button

indykoning:
I added two config values to allow a `pressed_action` and `release_action` to be added to allow this button to be used as a momentary button, calling a service eg: light.turn_on when it is pressed down and calling another service eg: light.turn_off to be called once this button is released.

PR #1014, opened 2025-09-30
feat(actions): press and release support for momentary actions

### LINKED ISSUE #249: Action of stop holding the button
Can we trigger a service or other action when the button is no longer pressed?
If not, are there any pointers on implementing this, so I can make a PR?

RomRider:
Supersedes #360 
Closes #249

### DIFF
--- a/src/action-handler.ts
+++ b/src/action-handler.ts
@@ -148,6 +148,10 @@ class ActionHandlerType extends HTMLElement implements ActionHandlerType {
       }
 
       if (options.isMomentary) {
+        // ignore anything but left mouse button
+        if ((ev as any).button !== 0) {
+          return;
+        }
         fireEvent(element, 'action', { action: 'press' });
         return;
       }


RomRider:
Didn't implement `repeat` support but is it even required ? I can't think of a use case...

dcapslock:
> Didn't implement `repeat` support but is it even required ? I can't think of a use case...

I can't think of a use case.

dcapslock:
Tested PR and works well. One quirk, maybe related to this PR or just actions in general, is that when you are using icon press/release, when you press, then move off icon to the button and release, the release action is fired. But if you release outside the button the action is not fired. Not sure if an issue or not, in that should the icon release be fired if not over the icon any more.

Code all looks good.

CubieMedia:
Works as intended from PR #360

Thank you @RomRider  for setting this up way better!

RomRider:
@dcapslock, it's the same behavior as with other actions: clicking on the icon and releasing on the card will fire the card `tap_action`. Clicking on the icon or on the card and releasing outside of the card will not do anything.
For momentary actions it could be problematic but I don't really see a way to handle that as we don't catch the release event on the proper element.

I'll merge as-is, and we'll see if we have complaints.

Issue #1021, opened 2025-10-04
Services are executed simply by switching to the dashboard view where the button-card is located, without clicking on the button


il77781:
**Checklist**
- [x] I updated the card to the latest version available
- [x] I cleared the cache of my browser

**Describe the bug**
Services are executed simply by switching to the dashboard view where the `button-card` is located (and sometimes, just by switching to any view of this dashboard that doesn't even contain this `button-card`), without clicking on the button.
**As far as I've noticed, this happens on cards that use a hack for multiple actions.** But of course, this is not certain...

**Version of the card**
Version: 5.0.0-dev.7 (perhaps also 5.0.0-dev.6)

**To Reproduce**
This is the configuration I used:
I noticed this on a complex `button-card` with several nested `button-cards` (and there `[[[[ - ]]]]` were used), but here I'm providing an example of a simple `button-card` where the same behavior occurs.
```yaml
type: custom:button-card
icon: mdi:weather-windy
name: " [[[ return states['fan.dmaker_p44_7a6c_fan'].attributes.preset_modes[0] ]]] "
tap_action:
  action: two_actions
  multi_calls: |
    [[[
      hass.callService(
       "fan",
       "set_preset_mode",
       { entity_id: "fan.dmaker_p44_7a6c_fan", preset_mode: states['fan.dmaker_p44_7a6c_fan'].attributes.preset_modes[0] }
      );
      setTimeout(() => {
        hass.callService(
          "xiaomi_miot",
          "get_properties",
          { entity_id: "fan.dmaker_p44_7a6c_fan", update_entity: true, mapping: [{ siid: 2, piid: 3 }] }
        );
      }, 50);
    ]]]

```
**Expected behavior**
In earlier versions, services were executed only after directly clicking on the button, which was required. I would like to return to the previous behavior...

**Smartphone:**
Chrome WebView (Android companion app 2025.8.7 - Android 11)
Home Assistant 2025.10.1

RomRider:
That is expected behavior with the latest version.

Those hacks are not meant to be used and not officially supported and can break anytime.

You can achieve exactly what you are trying to achieve with a simple script in HA.

RomRider:
I'll introduce a different feature to be able to run Javascript on any action, officially.

Mariusthvdb:
not only happening on multi call or whatever, but on 'anything' ....  ;-)
```
type: custom:button-card
template: subview
tooltip: Naar Dashboard
icon: mdi:chevron-double-left
[...]

PR #1023, opened 2025-10-05
fix(actions): avoid executing invalid actions configurations

RomRider:
BREAKING CHANGE: `*_action` stricly follow what is allowed in the configuration of this card (see updated documentation). If you used some hacks, it might break. If those hacks were created to run javascript code, you can now use `action: javascript` instead.


dcapslock:
Good solution.

RomRider:
Related to https://github.com/custom-cards/button-card/issues/1021#issuecomment-3368571503

PR #1033, opened 2025-10-06
feat(actions): confirmation support for javascript, PIN & Password confirmation support

RomRider:
WIP

### DIFF
--- a/src/button-card.ts
+++ b/src/button-card.ts
@@ -24,7 +24,6 @@ import {
   ActionEventData,
   EvaluatedActionConfig,
   ActionConfig,
-  JavascriptActionConfig,
   NoActionConfig,
   ToggleActionConfig,
   MoreInfoActionConfig,
@@ -33,6 +32,9 @@ import {
   PerformActionActionConfig,
   AssistActionConfig,
   CustomActionConfig,
+  ActionCustomEvent,
+  JavascriptActionConfig,
+  CallServiceActionConfig,
 } from './types/types';
 import { actionHandler } from './action-handler';
 import {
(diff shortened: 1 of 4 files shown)


RomRider:
Compiler was complaining about typings even though they are good, that's why there's some...
`evaledActionConfig = evaledActionConfig as XXXXXActionConfig;` in `_evalActions`

src/button-card.ts
```diff
@@ -223,6 +237,8 @@ class ButtonCard extends LitElement {
     }
   }
 
+  private _setupActionsCallback(): void {}
```
RomRider, line 240: Need to remove that

src/button-card.ts
```diff
@@ -1762,30 +1797,50 @@ class ButtonCard extends LitElement {
...
       const config = this._config;
       if (!config) return;
       const action = ev.detail.action;
       const actionKey = `${action}_action`;
       if (this._isActionDoingSomething(this._stateObj, config[actionKey])) {
-        this._runAction(ev.detail.action, { isIcon: false });
+        this._buildActionConfig(actionKey);
       }
     }
   }
```
RomRider, line 1815: Possible to combine handleCard and handleIcon

RomRider (commented):
Will fix those 2 comments tomorrow

src/button-card.ts
```diff
@@ -1601,44 +1617,53 @@ class ButtonCard extends LitElement {
...
 
   private _evalActions(config: ButtonCardConfig, action: string): ActionEventData {
-    const localActionEventData: ActionEventData = {};
+    let evaledActionConfig: EvaluatedActionConfig | undefined;
+    const staticAction = 'tap_action';
```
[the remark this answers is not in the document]
RomRider: Yes I thought about making it a const, makes sense. Will do.

PR #1041, opened 2025-10-07
feat(actions): call multiple actions in a row for any `*_action` or `icon_*_action` with the new `multi-actions` action

RomRider:
Multi actions support with delay (parsed by parseDuration)

```yaml
type: 'custom:button-card'
icon: mdi:console
name: multi-actions
tap_action:
  action: multi-actions
  actions:
    - action: call-service
      service: light.toggle
      service_data:
        entity_id: light.test_light
    - action: javascript
      javascript: |
        [[[
          this._sendToastMessage(`Waiting 3s...`);
        ]]]
    - delay: '[[[ return "3s" ]]]'
    - action: call-service
      service: light.toggle
      service_data:
        entity_id: light.test_light
```

### DIFF
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -163,13 +163,13 @@ export const styles = css`
   }
   @keyframes blink {
     0% {
-      opacity: 0;
-    }
-    50% {
       opacity: 1;
     }
-    100% {
+    50% {
       opacity: 0;
+    }
+    100% {
+      opacity: 1;
     }
   }
   @-webkit-keyframes rotating /* Safari and Chrome */ {


RomRider:
Side note: Should probably make a helper to send a toast message ^^

Mariusthvdb:
this is revolutionary!  ;-)
amazing you are on fire indeed
cant wait for the use cases  we can manage with this

RomRider:
TBH, the only real use case I see is to run `javascript`, `navigate` or `fire-dom-event` alongside a service call. Everything else could **and should** be handled by a script in HA.

And there are some limitations to this... If you have a long multi-actions `delay` and you navigate away from the page, the actions will stop unfortunately.

Mariusthvdb:
yes combining frontend actions with javascript, and backend actions in a script is what I meant by revolutionary.... having the 2 wait for each other as a magical bonus

> If you have a long multi-actions delay and you navigate away from the page, the actions will stop unfortunately.

ah, that indeed is a pity, hadn't figured that....

RomRider:
It won't wait for the script to finish to trigger followup actions as we don't get feedback from backend about when the script finishes its execution, no change there.

RomRider:
Adding spinner...

Will make it configurable in another PR.

https://github.com/user-attachments/assets/97d9f189-e539-4c6c-89d7-ce593da8c551

RomRider:
Adding something @Mariusthvdb will like: `wait_completion` to wait until template resolves to true to go to next step

```yaml
          - type: 'custom:button-card'
            icon: mdi:console
            name: MA script completion
            variables:
              delay: 3s
            tap_action:
              action: multi-actions
              actions:
                - action: call-service
                  service: light.toggle
                  service_data:
                    entity_id: light.test_light
                - action: perform-action
                  perform_action: script.turn_on
                  target:
                    entity_id: script.delay_script
                - wait_completion: '[[[ return states["script.delay_script"].state === "off" ]]]'
                - action: call-service
                  service: light.toggle
                  service_data:
                    entity_id: light.test_light
```

RomRider:
I'll release, let's gather feedback :smile:

Mariusthvdb:
```
            variables:
              delay: 3s
```

aren't used are they.... maybe fix that example in that case, unless you had something else in mind ?

RomRider:
True, will fix that, thanks

src/styles.ts
Mariusthvdb, line 1: not exactly sure what is blinking, but do we set to 0 at 50%, or just a bit more, like 0.2.
personally I most of the time like that just a bit better, so the element is visible at all times, even if very faint

Mariusthvdb (commented):
sorry, meant to post as comment, not review...

PR #1064, opened 2025-10-15
fix: custom_fields cards are not recreated from scratch anymore if their config is modified

### LINKED ISSUE #1060: The spinner display and multi-actions are interrupted prematurely
**Checklist**

**Describe the bug**
I don't think this is a bug, but I've encountered this: after tapping the button, the spinner is displayed for a time that is even shorter than the time specified in the `delay` parameter (between actions). Consequently, the time for "locking" the button from subsequent taps is also reduced, and the actions specified after the `delay` parameter are not executed at all... It feels like the `multi-actions` are interrupted or "reset"... At the same time, of course, it is necessary to say that I do not switch the page on which the `button-card` is located, nor do I do any other actions...

**Version of the card**
Version: 5.1.0-dev.3

**To Reproduce**

```yaml
type: custom:button-card
variables:
  cond35: " [[[ return (states['switch.dmaker_p44_7a6c_physical_control_locked'].attributes['physical_controls_locked'] == true) ? true : false ]]] "
  cond36: |
    [[[
      if ((states['switch.dmaker_p44_7a6c_physical_control_locked'].state == 'on' && !variables.cond35)
      || (states['switch.dmaker_p44_7a6c_physical_control_locked'].state == 'off' && variables.cond35))
        return true
      else return false
    ]]]
styles:
  grid:
    - grid-template-areas: "\"Blanc ChLock\" \"Blanc ChLock\""
    - grid-template-columns: 1fr 1fr
    - grid-template-rows: auto
    - gap: 10px
custom_fields:
  Blanc:
    card:
      type: custom:button-card
      name: Blanc
  ChLock:
    card:
      type: custom:button-card
      entity: switch.dmaker_p44_7a6c_physical_control_locked
      name: Child lock
      styles:
        card:
          - opacity: " [[[ if (variables.cond35) return '0.4' ]]] "
        icon:
          - animation: " [[[ if (variables.cond36) return 'blink 1s linear infinite' ]]] "
      tap_action:
        action: multi-actions
        actions:
          - action: perform-action
            perform_action: switch.toggle
            target:
              entity_id: switch.dmaker_p44_7a6c_physical_control_locked
          - delay: 20s
          - action: perform-action
            perform_action: xiaomi_miot.get_properties
            data:
              entity_id: fan.dmaker_p44_7a6c_fan
              update_entity: true
              mapping:
                - siid: 7
                  piid: 1
      state:
        - value: "on"
          styles:
            card:
              - background-color: rgb(158, 158, 158)
              - opacity: " [[[ if (!variables.cond35) return '0.4' ]]] "
```
But this configuration is working correctly.:
```yaml
type: custom:button-card
styles:
  grid:
    - grid-template-areas: "\"Blanc ChLock\" \"Blanc ChLock\""
    - grid-template-columns: 1fr 1fr
    - grid-template-rows: auto
    - gap: 10px
custom_fields:
  Blanc:
    card:
      type: custom:button-card
      name: Blanc
  ChLock:
    card:
      type: custom:button-card
      entity: switch.dmaker_p44_7a6c_physical_control_locked
      name: Child lock
      styles:
        card:
          - opacity: " [[[[ if (states['switch.dmaker_p44_7a6c_physical_control_locked'].attributes['physical_controls_locked'] == true) return '0.4' ]]]] "
        icon:
          - animation: |
              [[[[
                if ((states['switch.dmaker_p44_7a6c_physical_control_locked'].state == 'on' && (states['switch.dmaker_p44_7a6c_physical_control_locked'].attributes['physical_controls_locked'] == false))
                || (states['switch.dmaker_p44_7a6c_physical_control_locked'].state == 'off' && (states['switch.dmaker_p44_7a6c_physical_control_locked'].attributes['physical_controls_locked'] == true)))
                  return 'blink 1s linear infinite'
              ]]]]
      tap_action:
        action: multi-actions
        actions:
          - action: perform-action
            perform_action: switch.toggle
            target:
              entity_id: switch.dmaker_p44_7a6c_physical_control_locked
          - delay: 20s
          - action: perform-action
            perform_action: xiaomi_miot.get_properties
            data:
              entity_id: fan.dmaker_p44_7a6c_fan
              update_entity: true
              mapping:
                - siid: 7
                  piid: 1
      state:
        - value: "on"
          styles:
            card:
              - background-color: rgb(158, 158, 158)
              - opacity: " [[[[ if (states['switch.dmaker_p44_7a6c_physical_control_locked'].attributes['physical_controls_locked'] == false) return '0.4' ]]]] "
```

**Expected behavior**
Correct display of the spinner and execution of all `multi-actions` for the configuration specified in the first example

Chrome WebView (Android companion app 2025.8.7 - Android 11)
Home Assistant 2025.10.1

**Additional context**
The two configurations difference is that in the second configuration the JavaScript templates in the nested `button-card` are enclosed in the "correct" (corresponding to the level of nesting of the `button-card`) number of square brackets (`[[[[ ]]]]` instead of `[[[ ]]]` in the thirst example).
And if the specified `delay` time between actions is longer than the sensor (which is specified in the JavaScript template in the nested card) update interval, the spinner (and the "blocking" of the card from being tapped) is displayed until the state (or attribute value) of the sensor is updated. At this point, the nested `button-card` is probably "recalculated" and the `multi-actions` are reset - actions (after a `delay`) are not executed...
To be honest, I don't know if there's any way to do something about it... Or if there's even a need to do anything... Maybe, at the very least, it should be specified in the documentation that the number of square brackets for JavaScript templates in the nested cards **must be strictly followed**, at least when `multi-actions` are used...
At the same time, of course, I want to say that the "mismatch" in the number of square brackets with the level of nesting of the card works without any problems in all other cases... In my specific case, I used triple square brackets (instead of `[[[[ ]]]]`) for javascript templates in nested cards to access variables defined in the main ("parent") `button-card` (as far as I understand, there is currently no other way to do this, maybe [this](https://github.com/custom-cards/button-card/issues/1059#issue-3507118498) will solve the problem from a different angle)...

RomRider:
Fixes #1060

### DIFF
--- a/src/button-card.ts
+++ b/src/button-card.ts
@@ -424,6 +424,9 @@ class ButtonCard extends LitElement {
   }
 
   protected shouldUpdate(changedProps: PropertyValues): boolean {
+    if (changedProps.has('_config')) {
+      return true;
+    }
     if (this._config?.triggers_update === 'update_timer') {
       if (changedProps.has('_updateTimerMS')) {
         return true;
@@ -1016,10 +1019,18 @@ class ButtonCard extends LitElement {
         };
         let thing;
         if (!deepEqual(this._cardsConfig[key], cards[key])) {
-          thing = this._createCard(cards[key]);
-          thing.preview = this.preview;
-          this._cards[key] = thing;
-          this._cardsConfig[key] = copy(cards[key]);
+          if ((this._cardsConfig[key] as any)?.type === cards[key]?.type) {
+            // same type, different config
+            thing = this._cards[key];
+            thing.preview = this.preview;
+            this._cardsConfig[key] = copy(cards[key]);
+            thing.setConfig(cards[key]);
+          } else {
+            thing = this._createCard(cards[key]);
+            thing.preview = this.preview;
+            this._cards[key] = thing;
+            this._cardsConfig[key] = copy(cards[key]);
+          }
         } else {
           thing = this._cards[key];
         }
--- a/test/ui-lovelace.yaml
+++ b/test/ui-lovelace.yaml
@@ -2822,6 +2822,28 @@ views:
           }; ]]]
 
 
+      - type: custom:button-card
+        entity: switch.skylight
+        name: Nested should not be recreated
+        tap_action:
+          action: toggle
+        styles:
+          grid:
+            - grid-template-areas: '"i" "n" "nested"'
+            - grid-template-rows: auto auto auto
+        custom_fields:
+          nested:
+            card:
+              type: custom:button-card
+              name: '[[[ return `Main card state: ${entity.state}`; ]]]'
+              styles:
+                name:
+                  - color: '[[[ return entity.state === "on" ? "green" : "red"; ]]]'
+              tap_action:
+                action: multi-actions
+                actions:
+                  - delay: 10000
+
   - title: Grid
     type: sections
     sections:


dcapslock:
Looks good. This is how hui-card handles in Frontend, only recreating card if card type changes.

luismalves:
This fix actually broke updates for me in this situation @RomRider (I'm using decluterring card):
The JS template inside custom fields is no longer evaluated on entity state change.

```
   entity: "[[entity]]"
    name: "[[name]]"
    tap_action:
      action: more-info
    show_icon: false
    show_state: false
    show_name: false 
    styles: !include ../styles/style-card-timer.yaml
    template: "[[template]]"
    type: "custom:button-card"
    custom_fields: !include ../icons/ac-timer.yaml
``` 

And the custom field section is:
```
script-1:
  card:
    type: "custom:decluttering-card"
    template: script-timer-time-button
    variables:
      - template: script-turn-off
      - name: |
          [[[
            return entity && entity.state === "active" ? '+15min' : '30min';
          ]]]
      - entity: "[[entity]]"
      - duration: |
          [[[
            if (!entity || entity.state !== 'active') return '00:30:00';
            
            const finishesAt = entity.attributes.finishes_at;
            if (!finishesAt) return '00:30:00';
            
            // Calculate current remaining time
            const finishTime = new Date(finishesAt);
            const now = new Date();
            const remainingSeconds = Math.max(0, Math.floor((finishTime - now) / 1000));
            
            // Add 15 minutes (900 seconds)
            const newTotalSeconds = remainingSeconds + 900;
            
            const hours = Math.floor(newTotalSeconds / 3600);
            const minutes = Math.floor((newTotalSeconds % 3600) / 60);
            const seconds = newTotalSeconds % 60;
            
            return `${hours.toString().padStart(2, '0')}:${minutes.toString().padStart(2, '0')}:${seconds.toString().padStart(2, '0')}`;
          ]]]
```

And this card in custom_fields is another button-card:
```
script-timer-time-button:
  card:
    name: "[[name]]"
    entity: "[[entity]]"
    tap_action:
      action: call-service
      service: timer.start
      service_data:
        entity_id: "[[entity]]"
        duration: "[[duration]]"
    show_icon: false
    show_name: true
    show_state: false
    type: "custom:button-card"
```

The name should on the nested button-card should have changed and now it doesn't on v6.0.0. With 5.2.0 it works as expected

PR #1089, opened 2025-10-28
feat(variables): variables can depend on any other variable (no alphabetical dependency anymore)

RomRider:
BREAKING CHANGE: `variables` are only evaluated if they are "used" so hacks using variables to run javascript code during card init for eg. should read the updated documentation section. It is still possible with the `force_eval` variables configuration option.

```yaml
type: custom:button-card
entity: switch.skylight
update_timer: 1s
variables:
  always_evaled:
    value: '[[[ window.alwaysEvaled = `${new Date().getTime()}`; ]]]'
    force_eval: true
  never_evaled: '[[[ window.neverEvaled = `${new Date().getTime()}`; ]]]'
name: 'always should update every second,<br/>never should be unset'
show_label: true
label: |
  [[[
    return `always: ${window.alwaysEvaled || "not set"}
      <br/>never: ${window.neverEvaled || "not set"}`;
   ]]]
```

```yaml
type: custom:button-card
variables:
  never_evaled: '[[[ throw new Error("This variable should never be evaluated") ]]]'
  aa:
    value: '[[[ return "Test variables dependencies: OK" ]]]'
  test1: '[[[ return variables.aa ]]]'
name: '[[[ return variables.test1; ]]]'
```

### DIFF
--- a/docs/source/advanced/config-templates.md
+++ b/docs/source/advanced/config-templates.md
@@ -130,8 +130,6 @@ state:
 
 You can add variables to your templates and overload them in the instance of your button card. This lets you easily work with templates without the need to redefine everything for a small change.
 
-Variables can be of any type: number, object, string, function, ...
-
 An example below:
 
 ```yaml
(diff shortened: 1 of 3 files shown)


dcapslock:
Nice use of proxy. 👏  To really make variables adhere to standard programming principles, I suggest to proceed with not evaluating all variables always at render. Perhaps use `variables_global` whch is a list of variable names to always evaluate. Won't be much of a change to code, and yes a breaking change, but much better to understand and debug ongoing for both users and maintainers.

lisz8:
The elimination of alphabetical dependency is great! As for the second change, I'm not super enthusiastic but I was able to get along with it :)

PR #1092, opened 2025-10-30
doc: Add Browser Mod nested templates example

dcapslock:
Added an example demonstrating nested templates in Browser Mod with a custom button-card configuration.


docs/source/examples/js-templates.md
```diff
@@ -102,3 +102,53 @@ styles:
         return entity?.state === 'on' ? 'Light On' : 'Light Off';
       ]]]
     ```
+
+## Nested templates in Browser Mod
+
+This examples shows the effect of template nesting when included code may require the template **CODE** rather that the template **VALUE**. Here Browser Mod code uses a button-card. However it may also be applicable if the Browser Mod code used another card that also supported `[[[ ]]]` javascript templates.
+
+https://github.com/user-attachments/assets/b81f840b-9b35-4693-8e61-0864e3dcb232
```
RomRider: This should be a video/gif added to the repo in the image folder. MkDocs doesn't know what to do with this kind of link (unlike github).

You can check the result using `cd docs && mkdocs serve --livereload`

RomRider (changes requested):

RomRider:
Thanks 😊

docs/source/examples/js-templates.md
dcapslock: Done. Videos need to use video tag. I had trouble with relative path so made absolute.

PR #1152, opened 2026-01-30
fix: Fix custom element does not exist error which may arise under specific circumstances

### LINKED ISSUE #1150: Custom_Field cards issues
since upgrading to latest version of HA I am no longer able to use the standard HA 'conditional' card as a custom_field card.  I now get an error to say the card doesn't exist.

I now get an error where this use to work perfectly before;

Image

Image

dcapslock:
Fixes #1150 

This is hard to replicate. To do so I set the initial values of helpers to undefined and then delayed calling the helperPromise in _createCard. Even then it only happens for me 1 out of 10 or so.

This PR may not be the best way. The crux of the matter is that Frontend does not await import of lazy loaded stock cards like it does for dialogs. Perhaps createThing is not needed anymore and solely rely on createCardElement and only show 'Loading...' while cardHelpers are not defined, allowing the recovery logic in createCardElement to handle the lazy loading, which it does, including recovering from not awaiting the import.

The PR as is provides for least risk as far as changes to button-card. However I am happy to look for alternatives. Two in mind:

1. Use createCardElemenet only and warp in a promise to allow for cardHelpers not to be available, and use `${until(thing), "Loading..."}` or
2. Switch over to load cards via hui-card method where everything is abstracted to Frontend. e.g. as per my [update to auto-entities](https://github.com/thomasloven/lovelace-auto-entities/pull/591/changes#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR186). Similar method used with expander-card. The only caveat here is the extra `hui-card` wrapper which may break card-mod config users have, and perhaps other things I have not thought of.

PR in draft to allow for discussion.

### DIFF
--- a/src/button-card.ts
+++ b/src/button-card.ts
@@ -1113,7 +1113,18 @@ class ButtonCard extends LitElement {
           'grid-area': key,
         };
         let thing;
-        if (!deepEqual(this._cardsConfig[key], cards[key])) {
+        if (
+          (this._cards[key] as any)?.type === 'error' &&
+          (this._cards[key] as any)?.error?.includes("Element doesn't exist") &&
+          cards[key]?.type !== 'error' &&
+          customElements.get(cards[key]?.type)
+        ) {
+          // previously errored because element didn't exist, but now it does
+          thing = this._createCard(cards[key]);
+          thing.preview = this.preview;
+          this._cards[key] = thing;
+          this._cardsConfig[key] = copy(cards[key]);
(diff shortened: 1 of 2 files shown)


RomRider:
I don't really understand why this is happening (lack of enough javascript knowledge...), but if the proper way to fix this once and for all is to load cards through `hui-card`, I think we should go this way.

